PR #3, opened 2016-12-20
Don’t expand `any` expressions with nested or/in functions when all function arguments are subvariable ids

dmonroy:
When discussing with @alejandro-rivera we found that we can identify if arguments are subvariable ids by comparing with the ids of the subvariables before expanding the expression with nested **in** and **or** functions.

Now the expression is expanded only when one or more of the arguments doesn't match with the subvariables' ids, but when all arguments are subvariables then the expression maintains the any function and the references got transformed to urls.

@jjdelc Can you confirm that this is the expected behaviour?

See the test: https://github.com/Crunch-io/scrunch/compare/master...dmonroy:has-any-processing?expand=1#diff-f7d0b592b836e5fde287dc1fc1993e74R1485

### DIFF
--- a/scrunch/expressions.py
+++ b/scrunch/expressions.py
@@ -433,36 +433,60 @@ def process_expr(obj, ds):
                 subvariables = arrays[0]
                 value = values[0]
 
-                if op == 'all':
-                    if len(value['value']) != 1:
-                        raise ValueError
-                    value['value'] = value['value'][0]
+                def _extract_id(url):
+                    '''Extracts the id form the given url'''
+                    return [
+                        part for part in url.split('/') if part != ''
+                    ][-1]
 
-                if len(subvariables) == 1:
-                    obj['function'] = real_op
-                    obj['args'][0] = {'variable': subvariables[0]}
-                    obj['args'][1] = value
+                subvariables_dict = {
+                    _extract_id(subvar): subvar
+                    for subvar in subvariables
+                }
+
+                # no expansion is needed if all arguments are subvariables,
+                # instead we should just transform variable references to
+                # their corresponding urls
+                are_subvars = all(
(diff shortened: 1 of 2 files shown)


jjdelc:
This this is great, this is the behavior I need.

I made a note about peeking in the URLs to read resource information, it's a pattern that we should avoid in favor of using the Shoji API to get the attributes we need.

scrunch/expressions.py
```diff
@@ -433,36 +433,60 @@ def _process(obj, variables):
...
+                are_subvars = all(
+                    subvar in value['value']
+                    for subvar in subvariables_dict.keys()
+                )
 
-                if len(subvariables) == 1:
-                    obj['function'] = real_op
-                    obj['args'][0] = {'variable': subvariables[0]}
-                    obj['args'][1] = value
+                if are_subvars:
```
jjdelc: I will need to revisit this later, the way we'll be dealing with numeric subvariable responses is by using the following mask `"%s_%s"  % (parent_id, number)`, this will happen in Scrunch under the covers, I'll add the other case later when my branch lands.

scrunch/tests/test_expressions.py
```diff
@@ -1482,6 +1482,39 @@ def _session_get(*args, **kwargs):
...
+                },
+                {
+                    'value': [
+                        '/dataset/variables/url/a/subvariables/1/',
+                        '/dataset/variables/url/a/subvariables/2/'
+                    ]
+                }
+            ]
+        }
+
```
jjdelc: 👍

dmonroy:
I removed the `_extract_id(url)` that was using before to obtain the id form a subvariable url, now it is using the id from the `var.subvariables`.

PR #9, opened 2017-01-05
Use tox, flake8 and isort

dmonroy:
Don´t merge yet!

This effort is to make simple to run tests for all supported python version with a simple command and at the same time pass code quality checks using flake8 and isort.

- [x] pass tests on all supported python versions from tox
- [x] pass isort checks
- [x] pass flake8 checks
- [x] document tox usage


dmonroy:
There is no final place for storing development docs, so for the moment lets use this pull request to write about tox usage.

## Install

Tox is installed from the dev_requirements.txt file, the following command is enough to make it available:

```
pip insall -r dev_requirements.txt
```

## Use

We can run all tox enviroments at once by just executing the `tox` command, it will run all tests for all configured python versions plus flake8 and isort.
```
tox
...TRIMMED// progress of tests for each environment ... 
___________________ summary ___________________
  py27: commands succeeded
  py34: commands succeeded
  py35: commands succeeded
  py36: commands succeeded
  flake8: commands succeeded
  isort: commands succeeded
  congratulations :)
```

Running this requires us to have all python versions for each configured environment available on our development machine, but if we don't have one or more python versions we can still run tests for the other python versions or environments by explicitly passing environment name to the tox command with `-e ENV_NAME`, something like:

```
tox -e py35
```

All available tox environments are: `py27, py34, py35, py36, flake8, isort`

scrunch/expressions.py
```diff
@@ -241,10 +242,10 @@ def _parse(node, parent=None):
             else:
                 for _name, _val in fields:
                     if not isinstance(node, ast.UnaryOp) and (
-                            isinstance(_val, ast.BoolOp)
-                            or isinstance(_val, ast.UnaryOp)
-                            or isinstance(_val, ast.Compare)
-                            or isinstance(_val, ast.Call)):
+                            isinstance(_val, ast.BoolOp) or
+                            isinstance(_val, ast.UnaryOp) or
```
jjdelc: `isinstance()` supports tupples, you can replace this chain of `or`s with `isinstance(_val, (t1, t2, t3...))`

scrunch/expressions.py
```diff
@@ -305,8 +306,8 @@ def _parse(node, parent=None):
                             # For method calls, we only allow list-of-int
                             # parameters.
                             if _name == 'args' and func_type == 'method':
-                                if 'value' not in right \
-                                        or not isinstance(right['value'], list):
+                                if 'value' not in right or\
```
jjdelc: This could be replaced with `not isinstance(right.get('value'), list)` instead of the `OR`ed condition. If `value` is not in the dictionary will return `None` and the condition will remain.

scrunch/expressions.py
```diff
@@ -408,7 +409,7 @@ def process_expr(obj, ds):
 
     def ensure_category_ids(subitems, variables=variables):
         var_id = None
-        var_value = None
+        var_value = None  # noqa: F841
```
jjdelc: Do we need this annotation here?

dmonroy: no, also the var_value is not needed anymore

PR #11, opened 2017-01-12
Extend exclusion filters

skriems:
This PR enables the current exclude Dataset method to extend an existing exclusion filter.

usage:
`ds.exclude(<expression>, 'and'/'or')`
`ds.exclude(<expression>, extend='and'/'or')`

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1075,7 +1075,7 @@ class Dataset(object):
             # on the expr to get rid of eventually unneeded parantheses
             fmt = '({}) {} ({})'
             expr = prettify(parse_expr(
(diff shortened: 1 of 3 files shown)


nealrichardson:
Why add an "extend" argument? Why is this not achieved by something like `ds.exclude(ds.exclusion and <expression>)` or similar?

skriems:
afaik that's impossible because there's no operator overloading in python for `is`, `and`, `or`, `not`. I looked at `functools.singledispatch` but you still have the problem with the operator in the call.

If you know a way of achieving this let me know. I currently can't think of a better way as using `ds.exclude(<expression>, <operator>)`

PR #41, opened 2017-01-27
Implement variable Categories handling

jjdelc:


scrunch/datasets.py
```diff
@@ -1393,6 +1396,17 @@ def __repr__(self):
...
 
+    _categories = None
+
+    @property
+    def categories(self):
+        if self.resource.type not in self.CATEGORICAL_TYPES:
+            raise TypeError("Variable of type %s do not have categories" % self.resource.type)
+        if self._categories is None:
+            self._categories = CategoryList(self.resource)
+        print(id(self._categories))
```
xbito: Leftover print

jjdelc: Removed!

scrunch/datasets.py
```diff
@@ -1393,6 +1396,16 @@ def __repr__(self):
     def __str__(self):
         return self.name
 
+    _categories = None
+
+    @property
+    def categories(self):
+        if self.resource.type not in self.CATEGORICAL_TYPES:
```
xbito: This is one of the cases where I'm not sure if we should internally use self.type or if it is ok to use self.resource.type.

PR #46, opened 2017-01-31
Fix integration tests

dmonroy:
WIP

Make the integration tests honor the newer versions of scrunch (new rules and function signatures).


dmonroy:
closing to force a new travis run now that master is fixed for python versions 3.x

PR #50, opened 2017-02-03
Prevent editing categories on derived variables. Fix the previous ime…

jjdelc:
The previous implementation on `Category().edit` was wrong because the resource attributes are inside its `.body` and not straight on it. 

On top of that, the actual ticket, to prevent editing categories if the variable is derived (has a `derivation`)


xbito:
Tests are not python 3 compatible. Darwin, Sebastian can you fix and merge please?

PR #54, opened 2017-02-07
Protect attributes

### LINKED ISSUE #38: Protect writing to attributes that should be accessed through .edit or shouldn't be writable at all
The Dataset and Variable classes have a list of mutable and immutable attributes. You can access them as attributes to be read, like "dataset.name" but, right now it is also possible to do dataset.name = "Something", that should be forbidden since we don't actually save that operation to Crunch. The only way to edit those attributes is by doing a dataset.edit(name="Something else"). Immutable attributes are not editable even with that syntax.

So the idea would be to have a class from which Dataset, Variable, (and potentially Category and others) will inherit that would implement that protection by looking at the _MUTABLE_ATTRIBUTES and _IMMUTABLE_ATTRIBUTES attributes lists.

skriems:
resolves #38 for
- Dataset
- Variable
- Category

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1454,15 +1454,11 @@ class Variable(ReadOnly, object):
     def __str__(self):
         return self.name
 
-    _categories = None
-
     @property
     def categories(self):
         if self.resource.body['type'] not in self.CATEGORICAL_TYPES:
             raise TypeError("Variable of type %s do not have categories" % self.resource.body.type)
-        if self._categories is None:
-            self._categories = CategoryList(self.resource)
-        return self._categories
+        return CategoryList._from(self.resource)
 
     def hide(self):
         LOG.debug("HIDING")
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -7,7 +7,8 @@ import re
 
 import six
 
-from scrunch.helpers import abs_url, subvar_alias, download_file, case_expr
+from scrunch.helpers import (abs_url, subvar_alias, download_file,
+                             case_expr, ReadOnly)
 
 import pandas as pd
 
@@ -680,7 +681,7 @@ class DatasetSettings(dict):
     del __readonly__
 
 
-class Dataset(object):
+class Dataset(ReadOnly, object):
     """
     A pycrunch.shoji.Entity wrapper that provides dataset-specific methods.
     """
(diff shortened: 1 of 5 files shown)


scrunch/datasets.py
```diff
@@ -680,7 +681,7 @@ def __readonly__(self, *args, **kwargs):
     del __readonly__
 
 
-class Dataset(object):
+class Dataset(ReadOnly, object):
```
jjdelc: We don't need `object` here anymore

scrunch/datasets.py
```diff
@@ -1456,15 +1454,11 @@ def __repr__(self):
...
-    _categories = None
-
     @property
     def categories(self):
         if self.resource.body['type'] not in self.CATEGORICAL_TYPES:
             raise TypeError("Variable of type %s do not have categories" % self.resource.body.type)
-        if self._categories is None:
-            self._categories = CategoryList(self.resource)
-        return self._categories
+        return CategoryList._from(self.resource)
```
jjdelc: Watch for a regression here, I had a test that made sure that you didn't return a new CategoryList instance each time you accessed the property. That `if` cached the instance once and reused it. With your change we always instantiate and init a new one. Is this deliberate?

scrunch/helpers.py
```diff
@@ -6,6 +6,26 @@
...
     from urllib.parse import urljoin
 
+
+class ReadOnly(object):
+    """
+    class for protecting undesired writes to attributes
+    """
+    def __init__(self, resource):
+        object.__setattr__(self, "resource", resource)
+        object.__setattr__(self, "session", resource.session)
```
jjdelc: Why do we store the .session and .url if we have the resource?

jjdelc:
Can you add some tests for this please? Maybe create a new ReadOnly subclass on the test file for this and poke at that, or use any of what we have just to make sure this is test covered.

Also, I made a note about a regression/test you removed from the CategoryList.

scrunch/datasets.py
skriems: the reason for that is related to the different approach of accessing the attributes with the new class. Formerly you were using the `fill` method which was directly setting the `Category` attributes (which should be forbidden by now). Changing their attributes with the `edit` method had no effect on the instances in the `CategoryList` so basically there was a conflict between having up to date `Category` instances in the `CategoryList` (`test_edit_category`) with a single intantiation of it (`test_instance_is_reused`).

scrunch/helpers.py
skriems: you're right. Will update and change the code where the session was accessed directly.

skriems:
Yes I was thinking this needs tests as well. Will update tomorrow.

PR #61, opened 2017-02-13
Feature/variable iterator

mathiasbc:


mathiasbc:
I left two FIXME notes in the code that need further work and to have tests passing. I am creating a pull request so you can review and comment what needs to be changed. Also, did not yet added tests for this feature since I don't know if it will be the final approach we will take.

I am not sure if it's a good idea to have Variable initialized by either a Shoji Tuple or a Shoji Entity. But I made it that way to have much of the code remain the same without the need to rewrite it.

Let me know what you think.

scrunch/datasets.py
```diff
@@ -680,7 +683,42 @@ def __readonly__(self, *args, **kwargs):
...
+        self.ds_resource = ds_resource
+        # self._vars = [Variable(var.entity, ds_resource) for var in ds_resource.variables.index.values()]
+        self._vars = self.ds_resource.variables.index.items()
+
+    def __iter__(self):
+        for var in self._vars:
+            yield var
+
+    def __len__(self):
+        return len(self.ds_resource.variables.index.keys())
```
jjdelc: Why is this not `len(self._vars)` ?

scrunch/datasets.py
```diff
@@ -680,7 +683,42 @@ def __readonly__(self, *args, **kwargs):
     del __readonly__
 
 
-class Dataset(object):
+class DatasetVariablesMixin(collections.Mapping):
```
jjdelc: Shouldn't `__getitem__` also be part of this `Mapping` mixin?

scrunch/datasets.py
```diff
@@ -1441,12 +1481,34 @@ class Variable(object):
...
-        self.resource = resource
-        self.url = self.resource.self
-        self.dataset = Dataset(dataset_resource)
+    def __init__(self, var_shoji, dataset):
+        self.shoji_tuple = var_shoji
+        self.is_instance = False
+        self._resource = None
+        self.url = None
+        # pass an Entity class to save time
+        if isinstance(dataset, pycrunch.shoji.Entity):
```
jjdelc: Do we have any case where we send the shoji entity on Variable init now? I ask so we don't unnecessarily support receiving the shoji entity and do hacks here leaving unused code.

mathiasbc: Yes, most cases do pass a shoji entity, we are only passing shijo tuple when doing ds.values(). If we only want to support passing a Shijo Tuple then we would need to rewrite all code referring to Variable.

scrunch/datasets.py
mathiasbc: I am not sure about this one. __getitem__ has logic that shouldn't be on a iterator in this case

scrunch/datasets.py
mathiasbc: yup, makes more sense

scrunch/datasets.py
```diff
@@ -1441,12 +1481,34 @@ class Variable(object):
...
+            self.dataset = dataset
+        else:
+            self.dataset = Dataset(dataset)
```
jjdelc: Why does the variable need a `Dataset` instance inside it? It should be able to make do with just the Shoji Entity, the `Dataset` instance should have helper methods that a scrunch user needs, but I don't think any operation that we need internally in the Variable instance, or do we? 

We'll be able to tell if all our usages are `self.dataset.resource`, then we don't need the instance.

scrunch/datasets.py
```diff
@@ -1441,12 +1481,34 @@ class Variable(object):
...
+    def resource(self):
+        if not self.is_instance:
+            self._resource = self.shoji_tuple.entity
+            self.url = self._resource.self
+            self.is_instance = True
+        return self._resource
 
     def __getattr__(self, item):
+        # don't access self.resource unless necessary
+        if item in self.shoji_tuple.keys():
```
jjdelc: here do `item in self.shoji_tuple` or use `hasattr`. No need to instantiate a list with all the items to check for attribute presence.

scrunch/datasets.py
```diff
@@ -1441,12 +1481,34 @@ class Variable(object):
...
+            return self.shoji_tuple[item]
         if item in self._ENTITY_ATTRIBUTES - self._OVERRIDDEN_ATTRIBUTES:
```
jjdelc: Shouldn't this check be before trying to access the tuple attribute?

scrunch/datasets.py
jjdelc: There isn't much logic that cannot be a try/except for the not found alias. 
https://github.com/Crunch-io/scrunch/blob/147f639bfc3a3cc5d28a08bb0d6d2db60d455c70/scrunch/datasets.py#L756

scrunch/datasets.py
mathiasbc: yes, it will equally work with self.dataset = dataset.entity instead of Dataset(dataset)

scrunch/datasets.py
mathiasbc: good point.

scrunch/datasets.py
mathiasbc: I don't get what you mean with that.

scrunch/datasets.py
mathiasbc: done.

scrunch/datasets.py
```diff
@@ -952,9 +987,9 @@ def create_single_response(self, categories,
                                  expr=expr,
                                  description=description))
 
-        return Variable(
-            self.resource.variables.create(payload).refresh(),
-            self.resource)
+        new_var = self.resource.variables.create(payload).refresh()
+        # return the variable instance
+        return self[new_var.body.alias]
```
jjdelc: This is not going to work. You have `self._vars` that reflects the old list of variables that does not contain this new variable there, so this will raise a variable not found error.
We need a hook that is part of the Mixin that refreshes the variable list. On the init we'd also call that same method and this will be what we use to invalidate cache.

scrunch/datasets.py
```diff
@@ -952,9 +987,9 @@ def create_single_response(self, categories,
                                  expr=expr,
                                  description=description))
 
-        return Variable(
-            self.resource.variables.create(payload).refresh(),
-            self.resource)
+        new_var = self.resource.variables.create(payload).refresh()
```
jjdelc: We can save ourselvs the `.refresh()` here since we are not using the shoji entity for anything we don't need to do a GET for a resource we are not using.

Unless make the Variable class accept an optional keyword argument for the full resource, that way we'd save ourselves a future GET trying to lazy fetch the attributes. 

So either way, remove the `.refresh()` calls on creations or have the Variable take an optional full entity resource to avoid it being lazy when we do have the full information.

scrunch/datasets.py
```diff
@@ -1115,7 +1152,7 @@ def combine_multiple_response(self, variable, map, categories=None, default=None
         responses into new categorized ones
 
             map={
-                1: 1,
+                1: [1],
```
jjdelc: why this? From what I understand, doug and users don't like to use lists where it isn't necessary.

if you ask me, I'd love to require single element lists for consistency sake, but I recall doug's example's never liking to do single element lists.

scrunch/datasets.py
```diff
@@ -1441,12 +1480,32 @@ class Variable(object):
...
+    def __init__(self, var_tuple, dataset):
+        self.shoji_tuple = var_tuple
+        self.is_instance = False
+        self._resource = None
+        # pass an Entity class to save time
+        if isinstance(dataset, pycrunch.shoji.Entity):
+            self.dataset = dataset
+        else:
+            # Need to instantiate Dataset for ds.order method
+            self.dataset = Dataset(dataset)
```
jjdelc: Why do we need the `Dataset` class here? Why can we not just proceed with the dataset shoji resource?

The Dataset class has many user facing methods that we don't need here, we should be able to proceed with direct shoji operations to the resource from the internal implementation of Variable here

jjdelc: Additionally, we should avoid taking one type or the other, that will make us code inconsistently passing diffferent things in different places. IF we need to support two types it should be explicit on the call, not magicly resolved.

scrunch/datasets.py
```diff
@@ -1441,12 +1480,32 @@ class Variable(object):
...
+    @property
+    def resource(self):
+        if not self.is_instance:
+            self._resource = self.shoji_tuple.entity
+            self.is_instance = True
+        return self._resource
+
+    @property
+    def url(self):
+        return self.resource.self
```
jjdelc: You don't need to have the resource hydrated to get the url. You can set the `self.url` on init by just getting it from `tuple.entity_url` and save yourself a GET request unnecessarily here.

scrunch/datasets.py
```diff
@@ -1441,12 +1480,32 @@ class Variable(object):
...
 
     def __getattr__(self, item):
+        # don't access self.resource unless necessary
+        if hasattr(self.shoji_tuple, item):
```
jjdelc: This should go under the attribute check verification, else we risk to expose attributes we don't want.

scrunch/tests/test_recodes.py
```diff
@@ -127,6 +126,7 @@ def test_combine_categories_unknown_alias(self):
         assert 'Dataset mocked_dataset has no variable unknown' in str(err.value)
 
     def test_combine_categories_from_alias(self):
+        # PIZZA
```
jjdelc: ?

scrunch/datasets.py
```diff
@@ -680,7 +680,51 @@ def __readonly__(self, *args, **kwargs):
...
 
-class Dataset(object):
+class DatasetVariablesMixin(collections.Mapping):
+    """
+    Handles dataset variable iteration
+    """
+
+    def __init__(self, ds_resource):
+        self.ds_resource = ds_resource
+        self._vars = self.ds_resource.variables.index.items()
```
jjdelc: Given the comments below, this assignation should be inside a method `self._reload_variables()` that gets called here and also after each variable creation.

We should also store the `.by('alias')` there so we don't need to fetch the variable catalog each time we want to access a variable.

scrunch/datasets.py
```diff
@@ -680,7 +680,51 @@ def __readonly__(self, *args, **kwargs):
...
+
+    def __getitem__(self, item):
+        # Check if the attribute corresponds to a variable alias
+        variable = self.ds_resource.variables.by('alias').get(item)
```
jjdelc: Following the comment above, we should have this cached so we don't do a GET to the variables catalog (which can be huge) each time we want to read a variable. Else we're fetching the same response over and over

scrunch/datasets.py
mathiasbc: impulse

scrunch/tests/test_recodes.py
mathiasbc: placeholders from vim

scrunch/datasets.py
mathiasbc: Yeah, good point

scrunch/datasets.py
mathiasbc: We need an instance for this particular line: https://github.com/Crunch-io/scrunch/blob/master/scrunch/datasets.py#L1773. It's also not something I added. I was an instance all along.

scrunch/datasets.py
mathiasbc: +1

mathiasbc:
At this point all tests pass, but a recent change to master makes it difficult to merge with it.

Issue #59, opened 2017-02-10
Give a summarized version of the data


xbito:
Our cousin rcrunch has ways to present to the user a summary of the data (and some other information too):

```
[crunch] R> table(ds$profile_gender)
profile_gender
    Male   Female 
133635.5 139855.1 

[crunch] R> ds$profile_gender
Gender (categorical)
Are you male or female? (profile_gender)

          Count
Female 139855.1
Male   133635.5
```

We should also provide helpers for that.

jamesrkg:
I have a version of this that works well enough in a pinch, you might be able to use it to start work on this one:

```python
    def frequencies(ds, alias, subalias=None):
        """
        Return a dataframe of the frequencies for the given variable.

        Parameters
        ----------
        ds : scrunch.mutable_dataset.MutableDataset/scrunch.streaming_dataset.StreamingDataset
            The dataset to be queried.
        alias : str
            The alias of the variable being queried.
        subalias : str, default=None
            In the case of querying a single subvarable, the alias of that
            subvariable.

        Returns
        -------
        df : pandas.DataFrame
            The returned frequencies dataframe.
        """

        if subalias is None:
            var = ds[alias]
            if var.type in ['multiple_response', 'categorical_array']:
                df = pd.concat([
                    frequencies(alias, subvar.alias)
                    for subvar in var
                ], axis=1)
            else:
                df = pd.DataFrame(var.resource.frequencies['value'])[['id', 'count']]
                df.set_index('id', drop=True, inplace=True)
                df.columns = [alias]
        else:
            var = ds[alias][subalias]
            df = pd.DataFrame(var.resource.frequencies['value'])[['id', 'count']]
            df.set_index('id', drop=True, inplace=True)
            cols = pd.MultiIndex.from_product([[alias], [subalias]])
            df.columns = cols

        return df
```

xbito:
Hey Jj, what do you think about this one? Is there something in the API that we should use instead more directly, or is Jamie implementation good to go?

jamesrkg:
@xbito i think this ticket should pivot towards the integration of scrunch with crunch-cube.

I saw there is a `warsaw-demo` branch but we'll need to know what the roadmap for that integration is. @slobodan-ilic can you provide us with any insight?

slobodan-ilic:
@xbito @jamesrkg @jjdelc 
[...]

PR #74, opened 2017-02-27
summary method displaying cube views

skriems:
this PR introduced a proposal of a `dataset.get_cube` method for retrieving cube views alongside a `var.summary` method which produces basic output from it.

I left some `TODO`'s (also for discussion). I.e.: I'm not absolutely sure if and how we can/should simplify the usage of `dimensions` and `measures`. This simplification (especially for `cube_mean` with categoricals casted as `numerics`) should probably be handled by the summary method(?)

#59


scrunch/datasets.py
```diff
@@ -1882,3 +1989,56 @@ def move(self, path, position=-1):
             )
         target_group = self.dataset.order[str(path)]
         target_group.insert(self.alias, position=position)
+
+    def summary(self, names=False, tablefmt='psql'):
```
xbito: I'm not sure about allowing different formats here, maybe lets start with one default format (whatever you think is best) and then see if there is demand for others?

setup.py
```diff
@@ -45,6 +45,7 @@ def get_long_desc():
     install_requires=[
         'pycrunch',
         'six',
+        'tabulate',
```
xbito: How many dependencies does tabulate bring?

skriems: `extras_require` only lists [wcwidth](https://pypi.python.org/pypi/wcwidth)

scrunch/datasets.py
skriems: originally I was thinking it was a good idea to just pass over the possibility to change the output (i.e. writing some reStructured Text with tablefmt='rst'). Sure, this is an edge case. For displaying purpose I thaught the psql format would be nice because users might know the format from PostgreSQL.
But as @rdrivers mentioned, I could also just make it a pandas Dataframe.

scrunch/datasets.py
```diff
@@ -1572,6 +1573,112 @@ def create_categorical(self, categories, alias, name, multiple, description=''):
             return self.create_single_response(
                 categories, alias=alias, name=name, description=description)
 
+    def get_cube(self, dimensions=[], measures=None, filter=[], weight=None):
```
jjdelc: Does this function need to be a method of the dataset? 

Could we move this into a `cubes.py` module? And if necessary by the python API then just do `def get_cube(): cubes.get_cube(self)`?

The datasets.py file is growing enormously with dozens of different concerns.

scrunch/datasets.py
```diff
@@ -1882,3 +1989,56 @@ def move(self, path, position=-1):
...
+        else:
+            refs = cube_result['dimensions'][0]['references']
+            categories = cube_result['dimensions'][0]['type']['categories']
+
+            table = []
+            for num, c in enumerate(categories):
+                table.append([c['name'], counts[num]])
+            headers = ['category', 'count']
+
+        print('\n      alias: {alias}, id: {id}\n'
```
jjdelc: You shouldn't print to stdout by default. Make the function receive a StringIO and write to that. Then you can test the output or you can export this to a file if desired.

jjdelc:
Needs test

scrunch/datasets.py
skriems: Hey jjdelc. Just saw your comments.

No of cause it doesn't. I just thaught it would make sense since we get a _crunch:cube_ of a _dataset_. But yep - the datasets module is getting quite big so I'll move into a cubes module.

scrunch/datasets.py
skriems: Right. Can't remember why I didn't switched it back to _return_ the output

PR #78, opened 2017-03-08
Adding subvariable iteration

mathiasbc:
This feature is an approach to solve:
https://yougov.fogbugz.com/f/cases/46610/Access-to-subvariables

The behaviour is documented at: https://docs.google.com/document/d/1tnk_kPzsyWgSbXAyi1m5v17LG3t6L0OuzUgPGiqJrUQ/edit#heading=h.91c810lm76al

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1593,7 +1593,47 @@ class Dataset(ReadOnly, DatasetVariablesMixin):
                 categories, alias=alias, name=name, description=description)
 
 
-class Variable(ReadOnly):
+class DatasetSubvariablesMixin(collections.Mapping):
+    """
+    Handles a variable subvariables iteration in a dict-like way
+    """
+
+    def __getitem__(self, item):
+        # Check if the attribute corresponds to a variable alias
+        subvariable = self.resource.subvariables.by('alias').get(item)
+        if subvariable is None:
+            # subvariable doesn't exists, must raise a ValueError
+            raise ValueError('Variable %s has no subvariable %s' % (
+                self.resource['name'], item))
+        # subvariable exists!, return the subvariable Instance
+        return Variable(subvariable, self)
+
+    def __iter__(self):
+        for var in self.resource.subvariables.index.items():
+            yield var
+
+    def __len__(self):
+        return len(self.resource.subvariables.index.items())
+
+    def itervalues(self):
+        for _, var_tuple in self.resource.subvariables.index.items():
+            yield Variable(var_tuple, self)
+
+    def iterkeys(self):
+        for var in self.resource.subvariables.index.items():
+            yield var[1].name
+
+    def keys(self):
+        return list(self.iterkeys())
+
+    def values(self):
+        return list(self.itervalues())
+
+    def items(self):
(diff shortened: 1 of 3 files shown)


scrunch/datasets.py
```diff
@@ -1593,7 +1593,47 @@ def create_categorical(self, categories, alias, name, multiple, description=''):
...
+                self.resource['name'], item))
+        # subvariable exists!, return the subvariable Instance
+        return Variable(subvariable, self)
+
+    def __iter__(self):
+        for var in self.resource.subvariables.index.items():
+            yield var
+
+    def __len__(self):
+        return len(self.resource.subvariables.index.items())
```
jjdelc: No need to do `.items()` here, you're generating a new list with the items when you only want the length, doing len on the index will do the len on the dict which will be the same result without creating the intermediate items list.

scrunch/datasets.py
```diff
@@ -1593,7 +1593,47 @@ def create_categorical(self, categories, alias, name, multiple, description=''):
...
+        if subvariable is None:
+            # subvariable doesn't exists, must raise a ValueError
+            raise ValueError('Variable %s has no subvariable %s' % (
+                self.resource['name'], item))
+        # subvariable exists!, return the subvariable Instance
+        return Variable(subvariable, self)
+
+    def __iter__(self):
+        for var in self.resource.subvariables.index.items():
+            yield var
```
jjdelc: The detault iteration on a dict like object is for the keys, so here you should iter over iterkeys() and return those. I think you could just do `yield from self.resource.subvariables.index`

scrunch/datasets.py
```diff
@@ -1593,7 +1593,47 @@ def create_categorical(self, categories, alias, name, multiple, description=''):
                 categories, alias=alias, name=name, description=description)
 
 
-class Variable(ReadOnly):
+class DatasetSubvariablesMixin(collections.Mapping):
```
jjdelc: Could this subclass the other Mapping done for Variables? Why not?

scrunch/datasets.py
```diff
@@ -1593,7 +1593,47 @@ def create_categorical(self, categories, alias, name, multiple, description=''):
...
+    """
+    Handles a variable subvariables iteration in a dict-like way
+    """
+
+    def __getitem__(self, item):
+        # Check if the attribute corresponds to a variable alias
+        subvariable = self.resource.subvariables.by('alias').get(item)
+        if subvariable is None:
+            # subvariable doesn't exists, must raise a ValueError
+            raise ValueError('Variable %s has no subvariable %s' % (
```
jjdelc: This should be a KeyError

scrunch/datasets.py
jjdelc: Wait, here we want to iterate over the aliases, not the urls, then you could just return an iterator `(tup.name for tup in self.resource.subvariables.itervalues())`

mathiasbc: That doesn't work, It would have to be `yield from self.resource.subvariables.index.values()`

mathiasbc:
Great review @jjdelc. Should be all fixed now.

PR #83, opened 2017-03-23
Descriptive error messages for the expression parser

dmonroy:
_sorry for the typo in the branch name_.

I've added verbose messages for all exceptions thrown by the expression parser, except for one that I'm not able to test.

@xbito can you confirm messages are fine?


dmonroy:
All is covered now, should we merge?

PR #84, opened 2017-03-25
Filter class

mathiasbc:


scrunch/datasets.py
```diff
@@ -992,6 +1026,19 @@ def settings(self, _):
...
+        _filters = {}
+        for f in self.resource.filters.index.values():
+            filter_inst = Filter(f)
+            _filters[filter_inst.name] = filter_inst
+        return _filters
+
+    @filters.setter
+    def filters(self, _):
+        # Protect the `settings` property from external modifications.
+        raise TypeError('Use add_filter method or edit_filter to modify filters')
```
xbito: I don't see edit_filter.

scrunch/tests/test_datasets.py
```diff
@@ -3417,4 +3417,58 @@ def test_ds_keys(self):
...
+                    'function': '!=',
+                    'args': [
+                        {'variable': var.url},  # Crunch needs variable URLs!
+                        {'value': 0}
+                    ]
+                }
+            }
+        }
+        ds.resource.filters.create.assert_called_with(expected_payload)
+
```
xbito: Just for completeness I would add a test for the edit of one of the filter properties.

scrunch/datasets.py
mathiasbc: @xbito: edit_filter is on the Filter class. And is a general `edit()` that allows to edit name, expression and so on.

mathiasbc:
@xbito: All changes ready now

scrunch/datasets.py
xbito: Right, but the wording here suggest that there is an edit_filter method in the Dataset, which we don't

mathiasbc:
I see what you mean. Removed that part.

PR #91, opened 2017-04-18
Bugfix/subvariable iteration order

dmonroy:
when iterating the subvariables of a variable, the iterator now yields the subvariables ordered by id

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1879,11 +1879,9 @@ class DatasetSubvariablesMixin(DatasetVariablesMixin):
         """
         Iterable of subvariable aliases
         """
-        values = self.resource.subvariables.index.values()
-        sort_func = lambda x: x['id']
-
-        for svar in sorted(values, key=sort_func):
-            yield svar
+        sv_index = self.resource.subvariables.index
+        for sv in self.resource['body']['subvariables']:
+            yield sv_index[sv]
 
     def __len__(self):
         return len(self.resource.subvariables.index)
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1877,7 +1877,7 @@ class DatasetSubvariablesMixin(DatasetVariablesMixin):
 
     def __iter__(self):
         """
-        Iterable of subvariable aliases
+        Iterable of ordered subvariables
         """
         sv_index = self.resource.subvariables.index
         for sv in self.resource['body']['subvariables']:
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1879,7 +1879,10 @@ class DatasetSubvariablesMixin(DatasetVariablesMixin):
         """
         Iterable of subvariable aliases
         """
-        for svar in self.resource.subvariables.index.values():
+        values = self.resource.subvariables.index.values()
+        sort_func = lambda x: x['id']
+
+        for svar in sorted(values, key=sort_func):
             yield svar
 
     def __len__(self):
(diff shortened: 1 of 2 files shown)


jjdelc:
Do you want to always sort the variables by ID?

The subvariables have a user defined order when they are created. It is specified on the parent's .subvariables attribute. The subvariables should always follow that order and not by ID which may be different of what the user wanted when creating his array.

xbito:
@jjdelc there was a conversation that pointed us towards the unreliability of json objects ordering being the culprit on why it was being considered that subvariables were not being iterated in the correct order. Though I do realize that @dmonroy implementation is only relying on the fact that the example happened to have the categories ordered by id, not the fact that their objects were in a particular order.

@dmonroy can you think about a different implementation?

scrunch/datasets.py
```diff
@@ -1879,7 +1879,10 @@ def __iter__(self):
         """
         Iterable of subvariable aliases
         """
-        for svar in self.resource.subvariables.index.values():
+        values = self.resource.subvariables.index.values()
+        sort_func = lambda x: x['id']
+
+        for svar in sorted(values, key=sort_func):
```
xbito: I don't think that the intention of the ticket was to use a sorting mechanism. Instead we should be trying to serialize/de-serialize the objects in using ordered objects. We will just have to trust that on the other end of the serialize/de-serialize json dance they are doing the same.

dmonroy:
@xbito @jjdelc Let me try to describe the other method I'm thinking for this:

The variable entity has a _subvariables_ member, that is a list of subvariable urls:

``` v.resource['body']['subvariables']
['https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0001/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0002/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0003/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0004/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0005/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0006/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0007/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0008/',
 'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0009/']
``` 

It also has a _subvariables_ catalog that returns an object whose keys are the subvariable urls. Actually we are reading from that object, iterating over the values.

```v.resource.subvariables.index
{'https://app.crunch.io/api/datasets/aaa/variables/000522/subvariables/0001/': {'alias': 'sv_alias',
  'derived': False,
  'description': '',
  'discarded': False,
  'id': '0001',
  'name': 'SV Name',
  'notes': '',
  'scale': 'interval',
  'type': 'categorical'},
  ...
}
```
What we can do is (assuming the _subvariables_ member is an ordered list) iterate the list of urls in the _subvariables_ member of the variable instance and return the variable matching that url, something like this:

```[v.resource.subvariables.index[url] for url in v.resource['body']['subvariables']]
``` 

@jjdelc am I right assuming that the _subvariable_ member in the variable instance is ordered?

jjdelc:
@dmonroy correct, that is the order that you should respect.

scrunch/datasets.py
dmonroy: Now using the order from the elements in the subvariables member of the variable entity.

PR #100, opened 2017-04-26
CrunchBox support

skriems:
This PR introduces support the `CrunchBox`:
usage:
```
>>> var1 = ds['var1_alias']
>>> f = ds.filters['f1_name']
>>> box = ds.create_crunchbox('title', 'header', 'footer', 'notes', filters=[f, 'f2_name'], variables=[var1, 'var2_alias'])
>>> ds.crunchboxes
[<CrunchBox title="..." ...>, ...]
>>> box.id
...
>>> box.widget_url
https://...
>>> box.iframe(logo=None, dimensions=None)
<figure ...></figure><iframe...
>>> box.variables
[<Variables XY>, ...]
>>> box.filters
[<Filter XY, ...>, ...]
>>> ds.delete_crunchbox(**kwargs)  # i.e. ds.delete_crunchbox(id=box.id) or
>>> box.remove()
```

@jjdelc: the boxdata entity was a bit different; I hope the docstring and comments on the proposal help understand some of my decisions.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -997,8 +997,8 @@ class CrunchBox(object):
 
         def _figure(html):
             return '<figure style="text-align:left;" class="content-list-'\
-                    'component image">\n' + '  {}\n'.format(html) + \
-                    '</figure>\n'
+                    'component image">' + '  {}'.format(html) + \
+                    '</figure>'
 
         _iframe = (
             '<iframe src="{widget_url}" width="{dimensions[width]}" '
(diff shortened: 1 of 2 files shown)


skriems:
^ since I merged `master` into my branch we now have those commits here which apparently aren't on `dev`.

scrunch/datasets.py
```diff
@@ -880,6 +880,146 @@ def remove(self):
...
+            _iframe = _figure(_img) + _iframe
+
+        elif self.title:
+            _div = '<div style="padding-bottom: 12px">\n'\
+                   '    <span style="font-size: 18px; color: #444444;'\
+                   ' line-height: 1;">' + self.title + '</span>\n'\
+                   '  </div>'
+            _iframe = _figure(_div) + _iframe
+
+        print(_iframe.format(**locals()))
```
xbito: I would prefer having this return a string, and let the client decide if they want it printed out or do something else with it.

Now, this is a representation that is exclusive for the purposes of Yougov Daily right? I'm hesitant about adding html that is case specific, regarding the style tags, div styling, etc. Maybe we can provide the iframe part only with the style passed as a parameter and leave the rest for the Yougov Daily context?

skriems: I'm not sure if this is specific to YouGov Online - it' s litterally copy-pasted from the rcrunch repo but with the ability to change the dimensions.
But I don't like the print statement and especially the classes here as well. I was thinking about frameworks like Flask, jinja2 but they would print it internally I think.
I'll take a look next week

Issue #92, opened 2017-04-19
Add support for category order in create_categorical (multiple)


AlasdairE:
Currently the order of categories in a derived multiple response created using `create_categorical` is arbitrary. 
Could there be a re-ordering after variable creation based on the given order?

xbito:
@jjdelc Is that something that the API supports? How?

jjdelc:
Those are variables that use the `case` derivation expression. The arguments containing the conditions for each category are in the order the categories will have.

There is probably some order getting lost from the input of create_categorical to the way the final payload is constructed.

dmonroy:
I created a categorical using the following code:
```python
ostype = dataset.create_categorical([
            {'id': 1, 'name': 'UNIX', 'case': 'operating_system in ("FreeBSD", "MacOS", "Minix", "NetBSD", "Solaris")'},
            {'id': 2, 'name': 'GNU', 'case': 'operating_system == "Linux"'},
            {'id': 3, 'name': 'NT', 'case': 'operating_system == "Windows"'},
        ], alias='ostype', name='OS type', multiple=True)
```

Then, when I try to iterate to categories it returns the list of categories in order ([1,2])
```python
ipdb> ostype.categories
CategoryList([(1, Category(name=Selected, numeric_value=None, id=1, selected=True, missing=False)), (2, Category(name=Not selected, numeric_value=None, id=2, selected=False, missing=False))])
ipdb> [c for c in ostype.categories]
[1, 2]
``` 

I'm wondering if you are talking about the subvariables and not the categories, so checked it too:
```python
ostype.subvariables
['https://alpha.crunch.io/api/datasets/5cb5781a29494ceda831a2fb16536e53/variables/3d0cc0232fef438998416e3420c56b5f/subvariables/0001/',
 'https://alpha.crunch.io/api/datasets/5cb5781a29494ceda831a2fb16536e53/variables/3d0cc0232fef438998416e3420c56b5f/subvariables/0002/', 
 'https://alpha.crunch.io/api/datasets/5cb5781a29494ceda831a2fb16536e53/variables/3d0cc0232fef438998416e3420c56b5f/subvariables/0003/']

ipdb> [sv for sv in ostype]
[{'discarded': False, 'alias': 'ostype_1', 'name': 'UNIX', 'type': 'categorical', 'notes': '', 'derived': False, 'id': '0001', 'description': ''},
[...]

PR #105, opened 2017-05-09
Issue/92

dmonroy:
fixes the order issue described in #92


dmonroy:
should merge against dev, not master

PR #107, opened 2017-05-09
Update dev with changes from master

dmonroy:
there are changes in master that are not present in dev yet


dmonroy:
!excuse

this is attempting to merge dev into master, what we need is to merge master into dev

Issue #85, opened 2017-04-04
Support for the Crunch Decks API


jamesrkg:
Add helper/s to extend the tables/decks API to scrunch. A general overview of this would include:

* Build new table (assign side/top variables)
* Manage a deck (create, add to, remove from, save, share, etc.)
* Render a table as pandas DataFrame or a deck as a dict of pandas DataFrames
* Export decks to local XLSX

API reference for this is in Crunch: http://docs.crunch.io/#decks

We'll need API support for JSON output of deck contents — that isn't currently in the Crunch API.

mathiasbc:
To make it easier to track the progress of this: https://github.com/Crunch-io/scrunch/pull/110

mathiasbc:
A Dataset can have many Decks, created by the user and a Deck contains Slides (many) and a Slide finally can hold many Analyses. An Analysis in particular is metadata to generate a cube query. For the moment crunch only allows one Slide per Deck and one Analysis por Slide, so a Deck can be thought of as a direct cube query.

jjdelc:
Made a few comment on the PR

mathiasbc:
This is merged, Cube queries are on a different ticket.

PR #110, opened 2017-05-10
Feature/decks

mathiasbc:
Adds support for Decks iteration on a Dataset instance. Just putting the PR here so y'all can review it and give feedback. 

This is strictly for: https://github.com/Crunch-io/scrunch/issues/85

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1433,7 +1433,7 @@ class Dataset(ReadOnly, DatasetVariablesMixin):
                                  description=description,
                                  is_public=public))
         new_deck = self.resource.decks.create(payload)
-        return self.filters[new_deck.body['name']]
+        return self.decks[new_deck.body['name']]
 
     def create_single_response(self, categories, name, alias, description='',
                                missing=True, notes=''):
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -846,13 +846,14 @@ class Order(object):
         return self.graph[item]
 
 
-class Filter:
+class SubEntity(object):
     """
-    A pycrunch.shoji.Entity for Dataset filters
+    A pycrunch.shoji.Entity directly related to a Dataset.
+    For example; filters, decks
     """
-    _MUTABLE_ATTRIBUTES = {'name', 'expression', 'is_public', 'owner_id'}
-    _IMMUTABLE_ATTRIBUTES = {'id', }
-    _ENTITY_ATTRIBUTES = _MUTABLE_ATTRIBUTES | _IMMUTABLE_ATTRIBUTES
+    _MUTABLE_ATTRIBUTES = set()
+    _IMMUTABLE_ATTRIBUTES = set()
+    _ENTITY_ATTRIBUTES = set()
 
     def __init__(self, shoji_tuple):
         self.resource = shoji_tuple.entity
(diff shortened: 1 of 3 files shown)


mathiasbc:
from scrunch.datasets import get_dataset

ds = get_dataset(<dataset_id>)

# iterate decks on a dataset
for id, instance in ds.decks.items():
    print(instance)

# get an specific deck
my_deck = ds.decks[<deck_id>]

# iterate slides on a Deck
for id, instance in my_deck.slides.items():
    print(instance)

# get a particular slide
my_slide = my_deck.slides[<slide_id>]

# iterate analysis on a Slide
for id, instance in my_slide.analyses.items():
    print(instance)

# get a particular Analysis
my_analysis = my_slide.analyses[<analysis_id>]

# Ana analysis can be turned into a query cude
result = my_analysis.query_cube(ds)    # note to pass the dataset here

print(result.json())

scrunch/datasets.py
```diff
@@ -1394,6 +1374,14 @@ def add_filter(self, name, expr, public=False):
         new_filter = self.resource.filters.create(payload)
         return self.filters[new_filter.body['name']]
 
+    def add_deck(self, name, description="", public=False):
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 description=description,
+                                 is_public=public))
+        new_deck = self.resource.decks.create(payload)
+        return self.decks[new_deck.body['name']]
```
jjdelc: You're fetching the deck by name here, but under `.decks` they are keyed by `.id`. This will KeyError.
There is no constrain for deck names to be unique so it cannot be used as a key value.

scrunch/subentity.py
```diff
@@ -0,0 +1,142 @@
...
+    def __repr__(self):
+        return "<{}: id='{}'>".format(self.__class__.__name__, self.id)
+
+    def query_cube(self, ds):
+        """
+        :param: ds: Dataset() instance
+        out of the current instance GET a cube query
+        """
+        shoji_obj = self['query']
+        resp = ds.resource.session.get(
```
jjdelc: You don't need to do `.session.get` you can directly do `.resource.follow('cube', urlencode('query', json)))', that's the pycrunchy way to do it. No need to dig back to the `.session` to follow resources.

If you do `.session.get()` you're getting a `requests` response, you'll need to do `.payload` to get the value, which is digging inside the requests library to then dig out. Better to use `.follow()`

scrunch/tests/test_datasets.py
```diff
@@ -3591,3 +3592,47 @@ def test_filter_class(self):
         filter = MagicMock(entity=self._filter)
         mockfilter = Filter(filter)
         assert mockfilter
+
+
+class TestDeck(TestDatasetBase, TestCase):
```
jjdelc: Should test your `.decks` property accessor, it has a bug mentioned above when trying to fetch it by name.

scrunch/subentity.py
mathiasbc: Actually I followed the pycrunch way to do it: https://github.com/Crunch-io/pycrunch/blob/master/pycrunch/cubes.py#L70-L73

jjdelc: Then you'll need to also return `.payload` like pycrunch do it. But here I think the case is that pycrunch uses requests, and we use pycrunch. 

The situation I wanted to avoid is that I've seen many places in scrunch where in order to get a URL they're doing `resource.session.get(subresource_url)` where it is possible to follow `resource.subresource`. We should try to use pycrunch and not requests.

mathiasbc: Yes it's a better idea. I fixed this.

PR #118, opened 2017-05-16
feature/push rows

skriems:
same PR as the previous but now I derived my branch from dev...

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1338,13 +1338,14 @@ class Dataset(ReadOnly, DatasetVariablesMixin):
         the rows to appear in the dataset instead of waiting for crunch
         automatic batcher process.
         """
-        self.resource.batches.create({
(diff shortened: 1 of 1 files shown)


scrunch/datasets.py
```diff
@@ -1332,19 +1332,20 @@ def stream_rows(self, columns):
...
         automatic batcher process.
         """
-        self.resource.batches.create({
-            'element': 'shoji:entity',
-            'body': {
-                'stream': count,
-                'type': 'ldjson'
-            }
-        })
+        if bool(self.resource.stream.body.pending_messages):
```
jjdelc: We should provide a helper that creates the shoji wrappers. I see we're constructing shoji envelopes in many ways in different parts of the code. Something simple as `as_entity = lamda body: {"element": "shoji:entity", "body": body}` could do and give uniformity moving forward.

PR #121, opened 2017-05-18
Feature/mr in

mathiasbc:


jamesrkg:
I'll take a look, but first a question about the syntax. I can see from the tests that `'hobbies.any([1])'` is supported. Is `'hobbies in [1]'`?

mathiasbc:
@jamesrkg: It is not supported for multiple_response types. Perhaps I could expand to support it if you need it, I don't know if `in` is supported in crunch though, the problem is that I don't feel good modifying the parser to change `in` to `any` behind the scenes.

jamesrkg:
@mathiasbc that idea behind the ticket originally was for the back-end implementation of single vs multies to be invisible to the scrunch user writing expression syntax, so if `in` can be used for a `categorical` then the syntax supporting `multiple_response` should include `in` as well. Otherwise the user will need to remember which of these seemingly functionally identical approaches works for each variable, which is what I want to avoid.

Alternatively, deprecate `in` ... but that's a different topic of conversation.

mathiasbc:
It makes sense to make the interface more transparent to the user. I can't think of a reason to covert al `in` clauses to `any` behind the scenes; @jjdelc @xbito what do you think abou this ?

scrunch/expressions.py
```diff
@@ -475,20 +475,53 @@ def category_ids(var_id, var_value, variables=variables):
...
+            cat_to_ids = [
+                tup[0] for tup in aliases if int(tup[1].split('_')[-1]) in values]
+            return [{'variable': var_url}, {'column': cat_to_ids}], False
+
+        def is_multiple_response(var_url):
+            # make sure the variable is multiple response
+            var_id = variable_id(var_url)
+            for var in variables:
+                if variables[var]['id'] == var_id:
+                    if variables[var]['type'] == 'multiple_response':
```
jjdelc: Isn't the type available straight from the variable? `var_resource.body.type` instead of iterating? Consider that datasets can have multiple tens of thousands of variables, iterating over all variables should be considered an expensive operation.

scrunch/expressions.py
```diff
@@ -475,20 +475,53 @@ def category_ids(var_id, var_value, variables=variables):
...
             return value
 
+        def adapt_multiple_response(var_url, values):
+            # if variable is multiple_response
+            #       convert value --> column and change ids by aliases
+            var_id = variable_id(var_url)
+            aliases = []
+            for var in variables:
+                if variables[var]['id'] == var_id:
+                    for sub_var_id in variables[var]['subreferences']:
```
jjdelc: This info is available on the variable resource. On it you have `.subvariables` a list of ordered subvariables and `.subreferences` a dictionary keyed by subvariable url and the references for each so you can access `array_var_res.body.subreferences[suvariable_url]['alias']` in straight access

jjdelc:
yes that's what we should do for an expression `myvar in [1, 2, 3]` we should detect if `myvar` is MR or not and manipulate each type internally.

scrunch/expressions.py
mathiasbc: Yes, this is the proper way to do it. I just wanted to use the structure that we already have in the function, `variables` in this case is a structure that get's created in the function. But I will change to use the entity instead.

mathiasbc:
@jjdelc: should be ok to merge now.

scrunch/expressions.py
```diff
@@ -475,20 +489,37 @@ def category_ids(var_id, var_value, variables=variables):
...
+
         for item in subitems:
             if isinstance(item, dict) and 'variable' in item:
                 var_id = variable_id(item['variable'])
             elif isinstance(item, dict) and 'value' in item:
                 item['value'] = category_ids(var_id, item['value'])
             _subitems.append(item)
 
-        return _subitems
+        return _subitems, True
```
jjdelc: What is this True? Why is it needed?

If we're returning an extra parameter, update the function's docstring to indicate what it means.

mathiasbc:
I will need to add support for `==` and `!=`.

mathiasbc:
One thing we are missing, and I find difficult to implement is the operator `!=` which behind the scenes should be translated to `not varname all([])` but but having to negate the variable first is going to be troublesome since nesting is involved. Instead the same can be achieved with `not varname in []`.

jamesrkg:
@mbc I don't see `!=` being the same as `not all()` or `not in`. 

`!= [1, 2]` should mean _not equal to exactly `[1, 2]` but may have any other set of responses_. This could include `[2]`, `[1]`, `[1, 2, 3]`, `[1, 3]`, `[2, 3]`, `[3, 4]`, etc.. 

The reverse is true of `==` vs `in`. For example `== [1, 2]` should mean _equal to exactly `[1, 2]`, nothing more or less_ whereas `in [1, 2]` means _any of `[1, 2]` are in the responses, regardless of responses besides those_.

Since we've already established that `any()` is equivalent to `in`, `not any()` would be the same as `not in`.

This leaves `all()`/`not all()`. 

`all([1, 2])` should mean _has all of the codes `[1, 2]`, regardless of what other response/s it has_. This could include `[1, 2, 3]`, `[1, 2, 5, 7, 9]`, etc..

`not all([1, 2])` should mean _doesn't have all of `[1, 2]` but may have other responses_. This is almost the same as `!=` except that it disallows `[1, 2, 3]` because that includes all of `[1, 2]`.

xbito:
Jamie has made an excellent point which I didn't consider. == should not be translated to all then, since lets say if you have a multiple which has 1, 2 and 3 selected and you compare with an all(2, 3) would return True, but that's not an ==, so bottom line. I was wrong.

@jjdelc for that case, the ==, does crunch has an operation that applies to the multiple response case? Like an "only"? Or should we disallow those operations in scrunch expressions?

mbc:
hi! seems like you may have the wrong person here.

mathiasbc:
Yes Jamie figure that out yesterday, I didn't notice that either. I don't think there is a direct operator for translating `==`. We might need a combination of operators which is going to be painful for the expression processor.

mathiasbc:
I've deleted the transformation from `==` to `all`. We are currently only supporting syntax in the form `mr in [1,2,3]` and `mr not in [1,2,3]`. This should be enough for us to merge this PR and @xbito will open a new issue where we can discuss what to do for supporting the operator `==`.

PR #125, opened 2017-05-23
Support for get_datasets without the need to specify the project

mathiasbc:


scrunch/datasets.py
```diff
@@ -105,11 +105,10 @@ def _get_connection(file_path='crunch.ini'):
             "Unable to find crunch session, crunch.ini file or environment variables.")
 
 
-def get_dataset(dataset, connection=None, editor=False, project=None):
```
xbito: It wasnt' the intention of the ticket to remove the project parameter. We should keep it for the case when we pass a dataset by name and need a project to clarify which we are searching for (dataset names are not globally unique).

scrunch/tests/test_utilities.py
```diff
@@ -133,14 +133,10 @@ def test_get_dataset_from_project(self, session):
         projects.entity.datasets.by.side_effect = _by_side_effect(
             shoji_entity, ds_res)
 
-        ds = get_dataset('dataset_name', project='project_name')
```
xbito: We should also return this test to its previous state, and add a test for a get_dataset that passes a dataset id that is inside a project.

scrunch/datasets.py
```diff
@@ -127,22 +130,20 @@ def get_dataset(dataset, connection=None, editor=False, project=None):
...
+        shoji_ds = root.datasets.by('name')[dataset].entity
+    except KeyError:
         try:
-            shoji_ds = root.datasets.by('name')[dataset].entity
+            shoji_ds = root.datasets.by('id')[dataset].entity
         except KeyError:
             try:
-                shoji_ds = root.datasets.by('id')[dataset].entity
-            except KeyError:
+                dataset_url = '{}datasets/{}/'.format(root.self, dataset)
```
jjdelc: Better even, you should be able to do `root.catalogs.datasets` and only need to add the dataset id. That makes the URL guessing even smaller.
For URLs better to use something like `urljoin(root.catalogs.datasets, '{}/'.format(dataset))`

Issue #124, opened 2017-05-23
feature/mr_in breaks direct subvariable referencing syntax


jamesrkg:
The branch `feature/mr_in` throws an error on direct subvariable referencing syntax.

I've tried to reference a subvaraible by alias from both a `categorical_array` and a `multiple_response`, such as `'gornewGB_1 == 1'` where `gornewGB_1` is the alias of a subvariable within the variable `gornewGB`.

xbito:
So this must be fixed now?

PR #141, opened 2017-05-31
Guarding variable type check when referencing a subvariable from an e…

mathiasbc:
Solves: #124

### DIFF
--- a/scrunch/expressions.py
+++ b/scrunch/expressions.py
@@ -449,8 +449,9 @@ def adapt_multiple_response(ds, var_url, values):
     expression.
     """
     # convert value --> column and change ids to aliases
-    subvariable_urls = ds.variables.index[var_url]['subvariables']
-    variable = ds.variables.index[var_url].entity
+    var_index = ds.variables.index
+    subvariable_urls = var_index[var_url]['subvariables']
+    variable = var_index[var_url].entity
     aliases = []
     for sub_var in subvariable_urls:
         subvar_id = variable.subvariables.index[sub_var]['id']
@@ -526,7 +527,8 @@ def process_expr(obj, ds):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if var_url in ds.variables.index and ds.variables.index[var_url]['type'] == 'multiple_response':
+                var_index = ds.variables.index
+                if var_url in var_index and var_index[var_url]['type'] == 'multiple_response':
                     return adapt_multiple_response(ds, var_url, subitems[1]['value'])
 
         for item in subitems:
--- a/scrunch/expressions.py
+++ b/scrunch/expressions.py
@@ -526,9 +526,8 @@ def process_expr(obj, ds):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if var_url in ds.variables.index:
-                    if ds.variables.index[var_url]['type'] == 'multiple_response':
-                        return adapt_multiple_response(ds, var_url, subitems[1]['value'])
+                if var_url in ds.variables.index and ds.variables.index[var_url]['type'] == 'multiple_response':
+                    return adapt_multiple_response(ds, var_url, subitems[1]['value'])
 
         for item in subitems:
             if isinstance(item, dict) and 'variable' in item:
--- a/scrunch/expressions.py
+++ b/scrunch/expressions.py
@@ -526,8 +526,9 @@ def process_expr(obj, ds):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if ds.variables.index[var_url]['type'] == 'multiple_response':
-                    return adapt_multiple_response(ds, var_url, subitems[1]['value'])
+                if var_url in ds.variables.index:
+                    if ds.variables.index[var_url]['type'] == 'multiple_response':
+                        return adapt_multiple_response(ds, var_url, subitems[1]['value'])
 
         for item in subitems:
             if isinstance(item, dict) and 'variable' in item:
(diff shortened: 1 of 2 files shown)


scrunch/expressions.py
```diff
@@ -526,8 +526,9 @@ def category_ids(var_id, var_value, variables=variables):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if ds.variables.index[var_url]['type'] == 'multiple_response':
-                    return adapt_multiple_response(ds, var_url, subitems[1]['value'])
+                if var_url in ds.variables.index:
```
xbito: Why not an and instead of a nested if?

scrunch/expressions.py
```diff
@@ -526,7 +526,7 @@ def category_ids(var_id, var_value, variables=variables):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if ds.variables.index[var_url]['type'] == 'multiple_response':
+                if var_url in ds.variables.index and ds.variables.index[var_url]['type'] == 'multiple_response':
```
jjdelc: Careful with doing this. You probably fetched the catalog index before. This is doing 2 independent GET requests to /dataset/id/variables/ in the same line. It is better to first fetch `ds.variables.index` in one variable and then do these checks. Or even better, here https://github.com/Crunch-io/scrunch/blob/46737cd15f550c76a033844813eeee6652ff99f4/scrunch/expressions.py#L474 you fetched all the variables, probably that has the information you need for this condition without making this request.

I am ware of this request because the datasets that are being worked on can have around hundred thousands variables, so this catalog is not cheap to obtain.

mathiasbc: I will make a call just once, but I cannot use the value of `get_dataset_variables` because that  is a dictionary keyed by variable alias, I would have to iterate over items of the dict and match the one that has the variable url. It was like that in the past, and you commented and said that was scary too, because iterating a potential huge dictionary doesn't work either. So I will stick with callen `ds.variables.index` just once

scrunch/expressions.py
```diff
@@ -526,7 +527,8 @@ def category_ids(var_id, var_value, variables=variables):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if ds.variables.index[var_url]['type'] == 'multiple_response':
+                var_index = ds.variables.index
```
jjdelc: Note that you're still calling `.index` twice in the same code path first here and then inside inside `adapt_multiple_response` a few lines later here https://github.com/Crunch-io/scrunch/pull/141/files#diff-54e01706c90fd9d7e26c9f5f4d422b75R452.

The problem begins because you're using the variable URL to refer to it. You don't have the variable resource from before anywhere? It looks that we're reaching this point because somewhere before we could not get the variable shoji resource. You should be able to do `var_by_alias = ds.variables.by('alias')` at the top of the `process_expr` call and then use that inside all the loops you need instead of doing 2 GETs for each variable we want to validate in the expression. Then pass around that `var_by_alias` to the functions that need so you don't make more requests to get the same that you just got a few miliseconds ago.

After that change you will just be able to check if `variable.body.type == "multiple_response"` and avoid the double condition that looks non obvious (you're checking if the variable is in the index in case it is a subvariable).

You should avoid dealing with variable URLs as much as possible, I believe that most of this logic can be done without having to know the variable's URL,

scrunch/expressions.py
```diff
@@ -449,8 +449,9 @@ def adapt_multiple_response(ds, var_url, values):
...
     """
     # convert value --> column and change ids to aliases
-    subvariable_urls = ds.variables.index[var_url]['subvariables']
-    variable = ds.variables.index[var_url].entity
+    var_index = ds.variables.index
+    subvariable_urls = var_index[var_url]['subvariables']
+    variable = var_index[var_url].entity
     aliases = []
     for sub_var in subvariable_urls:
         subvar_id = variable.subvariables.index[sub_var]['id']
```
jjdelc: Note that here you're hitting the same `variable.subvariables` catalog each time inside the above iteration for each subvariable.
You can change this block like this:

```
for subvar_alias, subvar_tup in variable.subvariables.by('alias'):
  var_id = subvar_tup['id']
```

And you have all subvariables IDs and aliases in only one request. Even better, is it possible to know the same information fromt https://github.com/Crunch-io/scrunch/blob/1c92a46cbd713797c24f9cda85def7c78afea6e3/scrunch/expressions.py#L475 since it already has the subvariables information for the array variables https://github.com/Crunch-io/scrunch/blob/1c92a46cbd713797c24f9cda85def7c78afea6e3/scrunch/expressions.py#L430 this would avoid doing multiple requests per each variable in the expression.

scrunch/expressions.py
```diff
@@ -526,7 +527,8 @@ def category_ids(var_id, var_value, variables=variables):
         if len(subitems) == 2:
             if 'value' in subitems[1] and 'variable' in subitems[0]:
                 var_url = subitems[0]['variable']
-                if ds.variables.index[var_url]['type'] == 'multiple_response':
+                var_index = ds.variables.index
+                if var_url in var_index and var_index[var_url]['type'] == 'multiple_response':
                     return adapt_multiple_response(ds, var_url, subitems[1]['value'])
```
jjdelc: Note that your return value from `adapt_multiple_response` contains a variable url https://github.com/Crunch-io/scrunch/blob/1c92a46cbd713797c24f9cda85def7c78afea6e3/scrunch/expressions.py#L462 this probably means that you will use that to make another request to that URL to get that variable's attributes yet again, when you already had the variable shoji entity a few lines above. If you return the shoji entity, whoever receives that value can just read from that.

scrunch/expressions.py
mathiasbc: This is more complicated than that. In an ideal scenario I would know if a variable exists in a Dataset and just GET is once. But the expression parser knows nothing about this. It just transforms names to possible variable urls. At this step of the process it could be a variable that doesn't exist in the Dataset.

PR #143, opened 2017-06-02
122/create-group-location

skriems:
adds support for `before` and `after` arguments to the `Group.create_group` and `Group.move` methods. I opted for using the `insert` method two times in favor of not having similar code in two places. So we're sending now two `PUT's` (unfortunately?). What do you think @jjdelc?
I'm going to add tests asap.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -640,43 +640,20 @@ class Group(object):
 
         self.order.update()
 
-    def create_group(self, name, alias=None):
+    def create_group(self, name, alias=None, position=-1, before=None,
+                     after=None):
         name = self._validate_name_arg(name)
         elements = self._validate_alias_arg(alias)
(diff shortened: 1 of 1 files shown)


jjdelc:
@skriems I see on L651 and then on L655?

Why is it not possible to construct the groups in memory and then send the final payload to the server? Why is the intermediate PUT necessary? It looks like the `.insert()` method is doing two things:

1. Add the group to the in-memory variable
2. PUT to the server

And because of this implementation choice we're doomed to always write to the server each time we want to add a new group? If that's the case then it sounds like we need to split `.insert()` in two separate methods and call each when necessary not to make more server calls (and slower scripts) because of code debt.

skriems:
@jjdelc yes that makes sense. I'm going to create a card in trello for that and tackle it asap

Issue #151, opened 2017-07-31
Appending


xbito:
While we clarify if we are going to do a new library or a separate module/class to do appends we should start working on the code here at least. Let's do appends and also compares.

We should look into the possible difficulties on appends with primary keys, checking what is done in the R library  (If I remember correctly there was something done to primary keys so that appends behave as appends, nor as an upsert).

mathiasbc:
What exactly is an append and a compare ? I suppose appends are means of importing data into a Crunch dataset.

xbito:
https://github.com/Crunch-io/rcrunch/blob/master/R/append-dataset.R
https://github.com/Crunch-io/rcrunch/blob/master/R/compare-datasets.R

xbito:
Unsetting PK here: https://github.com/Crunch-io/rcrunch/blob/master/R/append-dataset.R#L37

jjdelc:
We should not do a separate library but we should implement it as different subclasses of a same BaseDataset class.

We should implement the StreamingDataset and MutableDataset classes, both inheriting from BaseDataset class.

All of our current Dataset methods would go either to BaseDataset or StreamingDataset. MutableDataset would implement methods that allow for schema changing operations.

All of our decks, filters, multitables, analysis, export operations should be made to take both subclasses and operate the same since they are oblivious to the streaming state of the dataset.

This is a great oportunity to rethink the way we develop the operations around a Dataset so we move away from the Dataset class being a bag with hundreds of every method possible and segregate the operations in other helper classes/functions.

We provide a /compare/ endpoint in the API that will simulate an append between two datasets and return the errors or success without really performing an append.

mathiasbc:
@jjdelc: Sounds good, I already implemented a compare that is a work in progress: https://github.com/Crunch-io/scrunch/blob/feature/material-joins/scrunch/append.py. I tried using the /compare endpoint but I only received 500 errors from Crunch. I asked Malecki on Slack but did not receive an answer. The documentation describes a GET with the appendable dataset url as a query parameter, are you sure this is a working feature ?

xbito:
I see there are now docs for another endpoint "align" https://github.com/Crunch-io/apidocs/pull/133/files which might be useful for this too.

I also see that there is a new "combination" feature, do we want to bring that over to scrunch too? https://github.com/Crunch-io/apidocs/pull/131/files

jjdelc:
[...]

PR #153, opened 2017-08-03
Feature/material joins

mathiasbc:
Related to https://github.com/Crunch-io/scrunch/issues/151


xbito:
The commits here don't seem to be related to material joins? And I thought that Crunch had a compare API request? Or is it correct that we do the comparison in scrunch code? @jjdelc could you take a look?

scrunch/datasets.py
```diff
@@ -2220,13 +2234,22 @@ def __iter__(self):
     def __len__(self):
         return len(self.resource.subvariables.index)
 
+    def variable_names(self):
+        """
+        Simply return a list of all subvariable names in the Dataset
+        """
+        sv_names = []
+        for var in self.resource.subvariables.index.items():
```
jjdelc: There seems to be a lot of overlapping between DatasetVariablesMixin and DatasetSubvariablesMixin but they don't have the right attribute stored.

If we stored `self._catalog` where we send `dataset.variables` for DatasetVariablesMixin and `variable.subvariables` for DatasetSubvariablesMixin then all these `variable_names`, `iterkeys` and iterator methods wouldn't need to be re implemented as they could just go `self._catalog.by('name').keys()`, and such without needing to do one implementation for `resource.variables` and another copy for `resource.subvariables`.

jjdelc:
Made a comment regarding the WETness of the DatasetSubvariablesMixin and DatasetVariablesMixin that can be simplified.

@mathiasbc let's discuss about the pros/cons of having different subclasses for our mutable and streaming datasets. My proposal is as follows:

```
                     BaseDataset
                      /                \
                   /                     \
   StreamingDataset    MutableDataset
```

Then keep the `get_dataset` function but have it return **only** StreamingDataset (we could add a warning and provide a `get_streaming_dataset` function later on) and have a parallel `get_mutable_dataset` (we can check with users if _mutable_ is the right term here, maybe "ended"?) that returns only MutableDataset, both would do the same but check on the attribute of the found dataset and raise an error if it doesn't match. 

My idea is that the scripts written with this have to make the explicit choice very early on about which kind of work they will be doing with the dataset.

mathiasbc:
Last commit makes a generic VariableMixin

mathiasbc:
@jjdelc: Just had a conversation with Fernando, I have now what I need to continue this separation of MutableDataset and StreamingDataset, but now it's not clear to me which methods should go on one class and which ones on the other as well and the methods shared by them.

mathiasbc:
@jjdelc: Last changes make the split of Dataset happen, please review the methods that BaseDataset, MutableDataset and StreamingDataset got and advice if any changes are needed. Basically this was a restructuring of the code.

scrunch/datasets.py
```diff
@@ -129,34 +113,23 @@ def get_dataset(dataset, connection=None, editor=False, project=None):
...
-                        "Dataset (name or id: %s) not found in context." % dataset)
-
-        ds = Dataset(shoji_ds)
-
-    if editor is True:
-        ds.change_editor(root.session.email)
-
-    return ds
+                dataset_url = urljoin(
+                    root.catalogs.datasets, '{}/'.format(dataset))
```
jjdelc: This case will never happend, this looks like the search by ID that you're doing above

scrunch/datasets.py
```diff
@@ -129,34 +113,23 @@ def get_dataset(dataset, connection=None, editor=False, project=None):
...
-            ds = project.get_dataset(dataset)
-    else:
+    # search by dataset name
+    try:
+        shoji_ds = root.datasets.by('name')[dataset].entity
+    except KeyError:
+        # search by dataset id
         try:
-            shoji_ds = root.datasets.by('name')[dataset].entity
+            shoji_ds = root.datasets.by('id')[dataset].entity
```
jjdelc: Better to factor out `dataset_catalog = root.datasets` and then do `dataset_catalog.by(...)` here and in L118, this way you only GET the datasets catalog once and not two times. If your catalog is big this will save you time.

scrunch/datasets.py
```diff
@@ -998,20 +937,27 @@ class DatasetVariablesMixin(collections.Mapping):
...
         Helper that takes care of updating self._vars on init and
         whenever the dataset adds a variable
         """
-        self._vars = self.resource.variables.index.items()
+        # this try except block is needed to update the 
+        # subvar or var catalogs once the Entity is modified
+        try:
+            self._catalog = self.resource.variables
+        except AttributeError:
+            self._catalog = self.resource.subvariables
```
jjdelc: This pattern smells like you will be better off with a subclass that knows which sub catalog to use and just re-implement this method to use `self.resource.subvariables`

scrunch/datasets.py
```diff
@@ -1025,8 +971,17 @@ def itervalues(self):
             yield Variable(var_tuple, self)
 
     def iterkeys(self):
+        """
+        Yield variable alias, since they are unique
+        """
         for var in self._vars:
```
jjdelc: If you have the `self._catalog` you could better do `return self._catalog.by('alias).iterkeys()`

scrunch/datasets.py
```diff
@@ -1025,8 +971,17 @@ def itervalues(self):
...
-            yield var[1].name
+            yield var[1].alias
+
+    def variable_names(self):
+        """
+        Simply return a list of all variable names in the Dataset
+        """
+        return [var[1].name for var in self._vars]
```
jjdelc: The same way, if you have the catalog resource, you can do `return self._catalog.by('name').iterkeys()`

scrunch/datasets.py
```diff
@@ -1038,10 +993,11 @@ def items(self):
         return zip(self.iterkeys(), self.itervalues())
```
jjdelc: `return self._catalog.by('alias').iteritems()`

scrunch/datasets.py
```diff
@@ -998,20 +937,27 @@ class DatasetVariablesMixin(collections.Mapping):
...
         whenever the dataset adds a variable
         """
-        self._vars = self.resource.variables.index.items()
+        # this try except block is needed to update the 
+        # subvar or var catalogs once the Entity is modified
+        try:
+            self._catalog = self.resource.variables
+        except AttributeError:
+            self._catalog = self.resource.subvariables
+        self._vars = self._catalog.index.items()
```
jjdelc: Looking at the usages below, looks like you don't need to store `self._vars`

scrunch/datasets.py
```diff
@@ -1088,6 +1044,47 @@ def editor(self, _):
...
+            who should be set as the new current editor of the given dataset.
+
+        :returns: None
+        """
+        def _host_from_url(url):
+            resolved = urlsplit(url)
+            return resolved.hostname
+
+        def _to_url(email):
+            api_users = 'https://{}/api/users/'.format(
```
jjdelc: You don't need to construct the URL like this. It looks like the bigger problem here is how to obtain the site's root when you only have the catalog resource. For now, make a module level function like this:
```
def root_from_resource(resource):
    return resource.session.get('https://{}/api/'.format(urlsplit(resource.self)))
```

scrunch/datasets.py
```diff
@@ -1088,6 +1044,47 @@ def editor(self, _):
...
+                _host_from_url(self.resource.self)
+            )
+            user_url = None
+
+            users = self.resource.session.get(api_users).payload['index']
```
jjdelc: Now that you have the root resource, if you have the email you can get the user like this:

```
user = root.users.by('email')[email]`
```

Then do `self.resource.edit(current_editor=user.entity_url)`

scrunch/datasets.py
```diff
@@ -1088,6 +1044,47 @@ def editor(self, _):
...
+                    self.resource.patch({'current_editor': url})
+                    break
+            assert user_url is not None, 'Unable to resolve user url'
+
+            return user_url
+
+        def _is_url(u):
+            return u.startswith('https://') or u.startswith('http://')
+
+        if isinstance(user, User):
```
jjdelc: If you have a user class, then you should move all this to that class and have this method receive a User entity and be it. Move all this logic to `User.get_user(email_or_url)`. All this logic does not belong to a `Dataset` class

This method's body would just check if it is a User, then do `self.resource.edit(current_editor=user.url)`, and if the incoming parameter is not a user, then do `user = User.get_user(user)` and then proceed.

scrunch/datasets.py
```diff
@@ -1768,103 +1285,14 @@ def savepoint_attributes(self, attrib):
                 'user_name'
                 'version'
         """
-
-        if len(self.resource.savepoints.index) == 0:
-            return []
-        else:
+        if len(self.resource.savepoints.index) != 0:
```
jjdelc: Factor out this `self.resource.savepoints` so when you use it again below you don't have to do the GET twice

scrunch/datasets.py
```diff
@@ -1768,103 +1285,14 @@ def savepoint_attributes(self, attrib):
...
             attribs = [
```
jjdelc: This list comprehension looks like `savepoints_resource.by(attrib).keys()`

scrunch/datasets.py
```diff
@@ -2034,275 +1454,67 @@ def export(self, path, format='csv', filter=None, variables=None,
...
-        """
-        sv_index = self.resource.subvariables.index
-        for sv in self.resource['body']['subvariables']:
-            yield sv_index[sv]
-
-    def __len__(self):
-        return len(self.resource.subvariables.index)
+    def get_exclusion(self):
+        # make sure there is an expression
+        expr = self.resource.exclusion.get('body').get('expression')
```
jjdelc: If `body` fails, then this will raise an AttributeError because will return None and None does not have a .get() method. Better to factor this logic to do it by steps.

```
exclusion = self.resource.exclusion
if 'body' not in exclusion:
    return None
expr = exclusion.get('expression')
return prettify(expr, self) if expr else None
```

Also does only 1 request in success case instead of 2

scrunch/datasets.py
```diff
@@ -2034,275 +1454,67 @@ def export(self, path, format='csv', filter=None, variables=None,
...
-    def itervalues(self):
-        for _, var_tuple in self.resource.subvariables.index.items():
-            yield Variable(var_tuple, self)
+    def add_filter(self, name, expr, public=False):
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 expression=process_expr(parse_expr(expr), self.resource),
+                                 is_public=public))
+        new_filter = self.resource.filters.create(payload)
+        return self.filters[new_filter.body['name']]
```
jjdelc: Looks like here you can save yourself all the iteration that happens inside `self.filters` and just `return Filter(new_filter.refresh())`. You already have the resource, no need to search for it by name again.

scrunch/datasets.py
```diff
@@ -2034,275 +1454,67 @@ def export(self, path, format='csv', filter=None, variables=None,
...
-    def iterkeys(self):
-        for var in self.resource.subvariables.index.items():
-            yield var[1].name
+    def add_deck(self, name, description="", public=False):
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 description=description,
+                                 is_public=public))
+        new_deck = self.resource.decks.create(payload)
+        return self.decks[new_deck.self.split('/')[-2]]
```
jjdelc: Same as filters, you can `return Deck(new_deck.refresh())` here and save yourself another iteration over the catalog and searching by ID.

Also, if you need the ID of a deck, you should have it under `new_deck.body.id`, it probably wasn't there to begin with because it was missing the `.refresh()` call that GETs the new resource to read all the new attributes.

scrunch/datasets.py
```diff
@@ -2034,275 +1454,67 @@ def export(self, path, format='csv', filter=None, variables=None,
...
+                                 expression=process_expr(parse_expr(expr), self.resource),
+                                 is_public=public))
+        new_filter = self.resource.filters.create(payload)
+        return self.filters[new_filter.body['name']]
 
-    def iterkeys(self):
-        for var in self.resource.subvariables.index.items():
-            yield var[1].name
+    def add_deck(self, name, description="", public=False):
+        payload = dict(element='shoji:entity',
```
jjdelc: We need a `shoji_entity` helper that does this wrapping for us, it is already being repeated here and I've seen it in many places: `payload = shoji_entity({name: ..., expression: ..., is_public:...})`

scrunch/datasets.py
```diff
@@ -2317,13 +1529,18 @@ class Variable(ReadOnly, DatasetSubvariablesMixin):
...
         """
         self.shoji_tuple = var_tuple
         self.is_instance = False
         self._resource = None
         self.url = var_tuple.entity_url
         self.dataset = dataset
+        # for Variables with subvariables
+        try:
+            self._reload_variables()
+        except AttributeError:
```
jjdelc: Why would this raise AttributeError? You were considering `.variables` and then `.subvariables` as 2nd case, if we have a 3rd case we didn't contemplate then it is programming error and should not pass silent.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+from scrunch.expressions import parse_expr, process_expr
+from scrunch.helpers import abs_url, case_expr, subvar_alias
+from scrunch.exceptions import InvalidDatasetTypeError
+from scrunch.datasets import (BaseDataset, _get_connection, _get_dataset,
+                              Variable, CrunchBox)
+from scrunch.subentity import Filter
+from scrunch.variables import (combinations_from_map, combine_categories_expr,
+                               combine_responses_expr, responses_from_map)
+
+_VARIABLE_PAYLOAD_TMPL = {
```
jjdelc: All usages of this template always overwrite the name/description/alias. This is yet another use of that `shoji_entity` helper mentioned above and there will be no need for this constant.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+    A simple wrapper of _get_dataset with streaming=False
+    """
+    shoji_ds, root = _get_dataset(dataset, connection, editor, streaming=False)
+    # make sure the Dataset is of type streaming != "streaming"
+    if shoji_ds['body'].get('streaming') == 'streaming':
+        raise InvalidDatasetTypeError("Dataset %s is of type 'streaming',\
+            use get_streaming_dataset method instead" % dataset)
+    ds = MutableDataset(shoji_ds)
+    if editor is True:
+        ds.change_editor(root.session.email)
```
jjdelc: If this is the case when we send the email to `change_editor` then better to factor out all that user searching logic out of `change_editor` and here do a `User.get_by_email()` and pass in that User entity. Will save lots of logic inside the Dataset class.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+def create_dataset(name, variables, connection=None):
+    if connection is None:
+        connection = _get_connection()
+        if not connection:
+            raise AttributeError(
+                "Authenticate first with scrunch.connect() or by providing "
+                "config/environment variables")
+
+    shoji_ds = connection.datasets.create({
+        'element': 'shoji:entity',
```
jjdelc: Another usage for the `shoji_entity` helper

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            }
+        }
+    }).refresh()
+    return MutableDataset(shoji_ds)
+
+
+class MutableDataset(BaseDataset):
+    """
+    Class that enclose mutable dataset methods or any
+    method that varies the state of the dataset and/or it's data.
```
jjdelc: There is a misunderstanding here, the Mutable dataset is about mutating the dataset's schema that will affect the streaming.

Changing dataset attributes is OK because it does not alter the schema. Changing dataset name, description, dates does not affect streaming and that should be allowed by the BaseDataset

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+
+        return self.resource.edit(**kwargs)
+
+    def delete(self):
+        """
+        Delete a dataset.
+        """
+        self.resource.delete()
+
+    def stream_rows(self, columns):
```
jjdelc: Move this method to the other dataset class. The way things will work is that if a dataset is accepting streaming then it won't accept normal appends, and vice versa.

If a dataset accepts streaming then its schema must be locked (read only dataset). So there will not be streaming support for a mutable dataset.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        if bool(self.resource.stream.body.pending_messages):
+            self.resource.batches.create({
+                'element': 'shoji:entity',
+                'body': {
+                    'stream': count,
+                    'type': 'ldjson'
+                }
+            })
+
+    def fork(self, description=None, name=None, is_published=False,
```
jjdelc: All datasets can be forked, this should go in BaseDataset

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            description = self.resource.body.description
+
+        body = dict(
+            name=name,
+            description=description,
+            is_published=is_published,
+            **kwargs
+        )
+        # not returning a dataset
+        _fork = self.resource.forks.create({"body": body}).refresh()
```
jjdelc: Use the shoji_entity helper here, the API will start becoming strict about expecting an `element` key.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        fork = MutableDataset(_fork)
+
+        if preserve_owner or '/api/projects/' in self.resource.body.owner:
```
jjdelc: You can set the new owner of the fork at creation time, refactor this logic to create a payload with the new owner on `.create()` and there will be no need to do this additional PATCH

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+                _fork.patch({'owner': self.resource.body.owner})
+                _fork.refresh()
+            except (pycrunch.ClientError, pycrunch.ServerError):
+                # Don't fail if the fork's owner cannot be set to be the same
+                # as the parent dataset.
+                pass
+
+        return fork
+
+    def create_multitable(self, name, template, is_public=False):
```
jjdelc: This method should go in BaseDataset

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            as_json['query'] = [parsed_q]
+            if 'transform' in q.keys():
+                as_json['transform'] = q['transform']
+            parsed_template.append(as_json)
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 is_public=is_public,
+                                 template=parsed_template))
+        new_multi = self.resource.multitables.create(payload)
+        return self.multitables[new_multi.body['name']]
```
jjdelc: Same as .filters and .decks, you can just return Multitable(new_multi.refresh()).

It looks like the .filters, .decks and .multitables properties are creating dicts that have new instances for all the objects in the catalog each time and most of times we access them only as dict to fetch just one of them, seems expensive.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        """
+        Copies a multitable from another Dataset into this one:
+        As described at http://docs.crunch.io/#post176
+        :name: Name of the new multitable
+        :multi: Multitable instance to clone into this Dataset
+        """
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 multitable=multi.resource.self))
+        self.resource.multitables.create(payload)
```
jjdelc: Receive the returned value here and just `return Multitable(new_multi.refresh())` without accessing `self.multitables` that is expensive (it GETs the catalog, creates new instances for each item and returns a dictionary) when we only need one of the instances and we already have it in memory here.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+
+        more_args = []
+        for case in cases:
+            more_args.append(parse_expr(case))
+
+        more_args = process_expr(more_args, self.resource)
+
+        expr = dict(function='case', args=args + more_args)
+
+        payload = dict(element='shoji:entity',
```
jjdelc: helper

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            resp_id = '%04d' % resp['id']
+            responses_map_ids.append(resp_id)
+            responses_map[resp_id] = case_expr(
+                case,
+                name=resp['name'],
+                alias='%s_%d' % (alias, resp['id'])
+            )
+
+        payload = {
+            'element': 'shoji:entity',
```
jjdelc: helper, will also save ton of typos

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+                })
+            return _subreferences
+
+        if variable.resource.body.get('derivation'):
+            # We are dealing with a derived variable, we want the derivation
+            # to be executed again instead of doing a `copy_variable`
+            derivation = abs_url(variable.resource.body['derivation'],
+                                 variable.resource.self)
+            derivation.pop('references', None)
+            payload = {
```
jjdelc: helper

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+                subreferences = subrefs(variable, alias)
+                for subref in subreferences:
+                    for subvar_pos in subvars:
+                        subvar = subvars[subvar_pos]
+                        if subvar['references']['name'] == subref['name']:
+                            subvar['references']['alias'] = subref['alias']
+                            break
+        else:
+            payload = {
+                'element': 'shoji:entity',
```
jjdelc: Lots of reuse of the same constant, use helper

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+                9: "no answer"
+            },
+            missing=9
+        """
+        if isinstance(variable, six.string_types):
+            variable = self[variable]
+
+        # TODO: Implement `default` parameter in Crunch API
+        combinations = combinations_from_map(map, categories or {}, missing or [])
+        payload = _VARIABLE_PAYLOAD_TMPL.copy()
```
jjdelc: Helper, no need for that template

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        if isinstance(variable, six.string_types):
+            parent_alias = variable
+            variable = self[variable]
+        else:
+            parent_alias = variable.alias
+
+        # TODO: Implement `default` parameter in Crunch API
+        responses = responses_from_map(variable, map, categories or {}, alias,
+                                       parent_alias)
+        payload = _VARIABLE_PAYLOAD_TMPL.copy()
```
jjdelc: same

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        else:
+            raise ValueError(
+                "Couldn't find a (unique) fork. "
+                "Please try again using its id")
+
+        body = dict(
+            dataset=fork_url,
+            autorollback=autorollback
+        )
+        self.resource.actions.post(body)
```
jjdelc: Maybe it is better to use .create() here, since this will be an async response, calling `.create()` will block and wait until it finishes. If we just use `.post` then this will continue and the next line in the script that called this may work assuming that `.merge()` called finished already while it is still processing.

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+                as_json['transform'] = q['transform']
+            parsed_template.append(as_json)
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 is_public=is_public,
+                                 template=parsed_template))
+        new_multi = self.resource.multitables.create(payload)
+        return self.multitables[new_multi.body['name']]
+
+    def import_multitable(self, name, multi):
```
jjdelc: Also in BaseDataset

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        :name: Name of the new multitable
+        :multi: Multitable instance to clone into this Dataset
+        """
+        payload = dict(element='shoji:entity',
+                       body=dict(name=name,
+                                 multitable=multi.resource.self))
+        self.resource.multitables.create(payload)
+        return self.multitables[name]
+
+    def create_single_response(self, categories, name, alias, description='',
```
jjdelc: The methods that create new derived variables (all of our current methods) can go in the BaseDataset class because they are always derived, they do not affect the schema that the streaming expects so their values always get re-evaluated without trouble. 

This and the following methods to create derived varibles should move to BaseDataset

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            'function': 'adapt',
+            'args': [
+                {'dataset': right_ds.url},
+                {'variable': right_var_url},
+                {'variable': left_var_url}
+            ]
+        }
+
+        # wrap the adapter method on a shoji and body entity
+        payload = {
```
jjdelc: helper

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        if multiple:
+            return self.create_multiple_response(
+                categories, alias=alias, name=name, description=description,
+                notes=notes)
+        else:
+            return self.create_single_response(
+                categories, alias=alias, name=name, description=description,
+                notes=notes)
+
+    def create_crunchbox(
```
jjdelc: Should to to BaseDataset, this doesn't change the schema

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            expr = process_expr(parse_expr(filter), right_ds)
+            payload['body']['filter'] = {'expression': expr}
+
+        progress = self.resource.variables.post(payload)
+        # poll for progress to finish or return the url to progress
+        if wait:
+            return wait_progress(r=progress, session=self.resource.session, entity=self)
+        return progress.json()['value']
+
+    def create_numeric(self, alias, name, derivation, description='', notes=''):
```
jjdelc: This and create_categorical, since they create derived variables, can go on BaseVariable

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+        """
+        Used to create new numeric variables using Crunchs's derived expressions
+        """
+        expr = parse_expr(derivation)
+
+        if not hasattr(self.resource, 'variables'):
+            self.resource.refresh()
+
+        payload = dict(
+            element='shoji:entity',
```
jjdelc: helper

scrunch/mutable_dataset.py
```diff
@@ -0,0 +1,892 @@
...
+            CrunchBox (instance)
+        """
+
+        if filters:
+            if not isinstance(filters, list):
+                raise TypeError('`filters` argument must be of type `list`')
+
+            # ensure we only have `Filter` instances
+            filters = [
+                f if isinstance(f, Filter) else self.filters[f]
```
jjdelc: I searched and looks like this is the only other case where we use `.filters`, then it looks like the property can go away and here just do `filters = [Filter(f) for f in self.resource.filters.itervalues()]`

scrunch/tests/test_datasets.py
```diff
@@ -163,9 +163,9 @@ def _get(*args):
 
 class TestDatasets(TestDatasetBase, TestCase):
 
-    def test_edit_Dataset(self):
+    def test_edit_dataset(self):
         ds_mock = self._dataset_mock()
-        ds = Dataset(ds_mock)
+        ds = MutableDataset(ds_mock)
```
jjdelc: When the methods mentioned in the comments above are moved to BaseDataset, here you will be able to use StreamingDataset and only use MutableDataset in the tests where they need that mutable method.

scrunch/datasets.py
```diff
@@ -129,34 +113,23 @@ def get_dataset(dataset, connection=None, editor=False, project=None):
...
+    # search by dataset name
+    try:
+        shoji_ds = root.datasets.by('name')[dataset].entity
+    except KeyError:
+        # search by dataset id
         try:
-            shoji_ds = root.datasets.by('name')[dataset].entity
+            shoji_ds = root.datasets.by('id')[dataset].entity
         except KeyError:
+            # search by id on any project
```
mathiasbc: This case builds the url and tries to find the given dataset within any project.

scrunch/datasets.py
mathiasbc: It will, if the dataset ID passed lives on a project. This was requested, so no need to specify projects anymore.

scrunch/datasets.py
mathiasbc: this won't work `self._catalog.by('alias')` if a JSONObject.

scrunch/datasets.py
mathiasbc: this won't work `self._catalog.by('name')` if a JSONObject.

scrunch/datasets.py
mathiasbc: this would return a tuple of (var_url, var_shoji) which is something we didn't want, we need (alias, Variable()). Also for the comments above we want the `self._vars` to save the `resource.index.items()` and access it just once; you suggested this some time ago when we first wrote this Mixin class

scrunch/datasets.py
mathiasbc: Subentities like Filter etc... are initialized with a ShojiTuple and not a ShojiEntity, this is so to avoid the roundtrip to Crunch and make it only if the user tries to access any attribute not present in the Tuple.

scrunch/datasets.py
mathiasbc: This is again the same issue as Filters, you suggested we instantiate Classes with ShojiTuple whenever possible, the refresh() method brings over an Entity.

scrunch/mutable_dataset.py
mathiasbc: Nice observation. I made it work directly with the `get_user` function, now takes all this logic out of Dataset.

scrunch/mutable_dataset.py
mathiasbc: but I think it's really just a dict of tuples and not an actual Entity, and I don't think a DS would have thousands of such subentities.

scrunch/mutable_dataset.py
mathiasbc: filters is a property that was requested, it's the only way for users to iterate over the filters of a dataset.

scrunch/helpers.py
```diff
@@ -107,3 +107,14 @@ def _validate_category_rules(categories, rules):
...
         )
+
+
+def shoji_wrapper(payload, shoji_class='entity'):
+    """
+    receives a dictionary and wraps its content on a body keyed dictionary
+    with the appropiate shoji:<class> attribute
+    """
+    return {
+        'element': 'shoji:{}'.format(shoji_class),
```
jjdelc: Note that if you pass `catalog` as value here, the final document will be wrong because it will include a `body`. Where catalogs need to have an `index`.

mathiasbc:
@xbito, @jjdelc: So this is approved it should be merged, it's going to make things harder for coming changes that I have for other issues.

xbito:
Merging the beast

PR #155, opened 2017-08-07
Order Hierarchy API for projects

### LINKED ISSUE #152: Order Hierarchy API within projects
Using rcrunch it's possible to create a project within another project, though I'm not sure this is possible using the web app.

Here's some code provided by Brian Law:

```r
# assume that 'ds' is the dataset you want to move
proj <- session()$projects$Discovery
datasets(proj) <- ds
ordering(datasets(proj))[['Omnibus projects']] <- union(entities(ordering(datasets(proj))[['Omnibus projects']]), self(ds))
```

We need access to this functionality, e.g.:

* Create a project within an existing project
* Manage sub-projects (owners, members, name, etc.)
* Get a dataset that's in a sub-project
* Move datasets to/from sub-projects

skriems:
This PR makes the order hierarchy api available to projects (#152). Since Datasets do not have an alias I'm using it's _ID_ instead.

```python
>>> Project.order
...
>>> Project.order['|'].create_group('FirstGroup', alias=[ds1.id, ds2.id], before..., after...)
>>> Project.order['|'].create_group('2ndGroup')
>>> ds1.move('|2ndGroup')
...
```
I noticed that the Crunch UI doesn't seem to reflect movements currently. These movements do happen on the `project.datasets.order` endpoint though, so I consider this as technically working.

Since the order hierarchy API in scrunch was quite bound to Datasets/Variables this PR makes some naming changes necessary to above methods I think. `alias` doesn't make much sense here and also some variable names in the code should be changed like `__migrate_var__` for instance. Should we start using a more generous name here like `elements` or `objects`, throwing a `DeprecationWarning` when `alias` is used (not harming functionality until next version)?

closes #152


xbito:
@jjdelc I'm curious if we are on the right path here trying the re-use the ordering classes/api that we implemented for variables to be used with projects/datasets. Do you have guidance on that?

scrunch/datasets.py
```diff
@@ -706,24 +731,34 @@ def move(self, path, position=-1, before=None, after=None):
...
+        self._datasets = None
         self._graph = None
         self._sync = True
         self._revision = None
 
     def _load_hier(self):
-        self._hier = self.ds.resource.session.get(
-            self.ds.resource.variables.orders.hier
-        ).payload
+        if hasattr(self.resource, 'variables'):
```
jjdelc: I don't think this switch will scale well to reuse the Order class.

All shoji orders are associated to a Shoji catalog. So instead of sending the dataset or project resource you should pass in the order resource and the catalog you are ordering, in the case of a dataset you pass the variables catalog. In the case of the project you pass the datasets catalog. That way you always have `self._catalog` that should match the items in the order regardless if they are variables, datasets or something else.

scrunch/datasets.py
```diff
@@ -761,15 +806,19 @@ def graph(self, _):
...
-                self.ds.resource.self + 'state/'
-            ).payload
-            if self._revision is None:
-                self._revision = ds_state.body.revision
-            elif self._revision != ds_state.body.revision:
-                # There's a new dataset revision. Reload the
-                # hierarchical order.
-                self._revision = ds_state.body.revision
+            if hasattr(self.resource, 'variables'):
+                ds_state = self.resource.session.get(
```
jjdelc: If we need different or extra rules for the variables order and the dataset order we should do this inside subclasses instead of cases here. Most of the operations should live in the `Order` class and if necessary subclass `DatasetVariableOrder`, `ProjectDatasetsOrder`. Hinting for attributes should be considered a bad practice

scrunch/datasets.py
jjdelc: I don't think this switch will scale well to reuse the Order class.

All shoji orders are associated to a Shoji catalog. So instead of sending the dataset or project resource you should pass in the order resource and the catalog you are ordering, in the case of a dataset you pass the variables catalog. In the case of the project you pass the datasets catalog. That way you always have `self._catalog` that should match the items in the order regardless if they are variables, datasets or something else.

scrunch/datasets.py
```diff
@@ -761,15 +806,19 @@ def graph(self, _):
...
+                    self.resource.self + 'state/'
+                ).payload
```
jjdelc: We don't concatenate strings, if you need the /state/ just follow `.resource.state`

scrunch/datasets.py
```diff
@@ -323,6 +324,17 @@ def edit_user(self, user, edit):
...
+
+    @order.setter
+    def order(self, _):
+        # Protect the `order` from external modifications.
+        raise TypeError(
+            'Unsupported operation on the Hierarchical Order property'
+        )
+
 
 class Path(object):
```
jjdelc: Can we move the Path and Order classes outside of datasets.py, maybe to orders.py? To remove bloat from the single file

scrunch/datasets.py
```diff
@@ -371,12 +383,23 @@ def __init__(self, obj, order, parent=None):
         # Load all the elements.
         for element in obj[self.name]:
             if isinstance(element, six.string_types):
-                _id = element.split('/')[-2]
-                var = self.order.vars.get(_id)
-                if var:
-                    self.elements[var['alias']] = var
+                if 'datasets' not in element or 'variables' in element:
```
jjdelc: Can you elaborate a bit more what is happening here?

In a shoji:order we should just be able to match the urls of the mentioned resources to the associated catalog. This looks like it is special casing in the parts of the URL to behave distinctly.

Alternatively, if this is a situation of relative vs absolute URLs, I think that pycrunch has some helpers to conver urls to relative/absolute, We should address that matching of urls with pycrunch and not here,

jjdelc:
I like that we're reusing the Path/Group/Order classes for the datasets order as well, but we should avoid doing switches for each kind. The shoji orders/catalog should map together regardless of their nature and if we need special casing we should then do subclasses for those bits only.

scrunch/datasets.py
skriems: Sure thing. `ds.variables.orders.hier` returns relative URL's to its variables while `project.datasets.order` returns absolute URL's. So I saw these 3 possibilities here:
- relative URLs to variables
- absolute URLs to variables (might change in the future)
- absolute URLs to datasets
- (touché, I missed relative URLs to datasets which could also change in the future)
I will take a look at pycrunch to handle this.

skriems:
@jjdelc I think I addressed all requests but two:
1. attaching the shoji.Order
2. using pycrunch.lemonpy.URL

Since we want to `self._sync` the `self._hier` property I haven't made up my mind howto achieve that without the resource. Probably a call to super or classmethod?

I was playing around a bit with the `URL` class in pycrunch but with no success so far since we have relative URL for variables (`../<var_id>/`) and absolute URL's for datasets which can't be relative to a project. I think we're also setting relative variable URL in the `_build_graph_structure` method:
https://github.com/Crunch-io/scrunch/pull/155/files#diff-10a14081413b0535e4d0097c2ad71a58L791

I assume we should avoid this since this violates the shoji spec?

scrunch/datasets.py
```diff
@@ -252,6 +252,7 @@ class Project:
     def __init__(self, project_resource):
         self.resource = project_resource
         self.url = self.resource.self
+        self._order = ProjectDatasetsOrder(self.resource)
```
jjdelc: Why is the project resource needed to init the order? I think that the init args here should be:
 * self.resource.datasets <- the datasets catalog
 * self.resource.datasets.order <- the order of the catalog we want this class to represent (because there may be more than one order)

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+                'Invalid path %s: cannot move Group into itself.' % path
+            )
+        target_group.insert(
+            self.name, position=position, before=before, after=after
+        )
+
+
+class Order(object):
+
+    def __init__(self, resource):
```
jjdelc: The order class shouldn't care too much about its parent resource (it may be necessary to instantiate variables, that should be handled in a subclass) but should primarily receive the Shoji:order and the associated Shoji:catalog. 
```
def __init__(self, order, catalog):
    self.+order = order
    self._catalog = catalog
```

Then in your subclasses you won't need to store specific `_vars` and `_datasets`, they are both the same `._catalog` resource you received.
```
class Dataset():
    ...
    def variable_order():   # Or however the order is obtained, not really important
        ds_resource = self.resource
        order = DatasetVariables(ds_resource.variables.hier, ds_resource.variables)
```

````
class Project():
    ...
    proj_resource = self.resource
    dataset_order = ProjectDatasets(proj_resource.datasets.order, proj_resource.datasets)
    ...
```

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+        self.resource = resource
+        self._hier = None
```
jjdelc: What is _hier and what is _graph?

I recommend storing _catalog and _order that map directly to the shoji:catalog and the shoji:order

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+        self._graph = None
+        self._sync = True
+
+    def _load_hier(self):
+        """
+        subclasses need to overwrite this method in order to get the updated
+        order hierarchy
+        """
+        return self._hier
```
jjdelc: Looks like _hier maps to the shoji:order resource? Why it needs a separate getter? You already receive the resource on init.

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+
+    def _load_graph(self):
+        self._graph = Group({'__root__': self.hier.graph}, order=self)
```
jjdelc: Looks like _graph is really your Scrunch Group class, maybe better name it `group` and no need for this to be lazy, you can just store `self.group = Group(self.order.graph)` on the `__init__` method.

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+        raise TypeError('Unsupported assignment operation')
+
+    @property
+    def graph(self):
+        if self._graph is None:
+            self._load_graph()
+        return self._graph
+
+    @graph.setter
+    def graph(self, _):
```
jjdelc: I think this is overkill, if we just set the `self.group = Group(...)` to be declared on init it will be a normal attribute, we cannot be protecting all attributes in python from being replaced, if a user is setting internal library attributes to something else, it will be their problem fro doing wrong usage. 

`self.order` and `self.group` don't need to be lazy nor to have `load_*` methods.

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+        self._hier = self.resource.session.get(
+            self.resource.variables.orders.hier).payload
+        return self._hier
+
+    def _load_vars(self):
+        self._vars = self.resource.variables.by('id')
+        return self._vars
+
+    @property
+    def vars(self):
```
jjdelc: We don't need this pattern here, if the DatasetVariablesOrder needs to have the variables by id, it can just happen on the subclass `__init__`:
```
def __init__():
    super()...
    self.variables = self.catalog.by('id')
```

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+        return self.graph[item]
+
+
+class DatasetVariablesOrder(Order):
+    def __init__(self, resource):
+        super(DatasetVariablesOrder, self).__init__(resource)
+        self._vars = None
+        self._revision = None
+
+    def _load_hier(self):
```
jjdelc: See comment above, rename `_hier` to store the shoji:order as `self.order` straight, no need for this lazy loading pattern

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+        return self._vars
+
+    @vars.setter
+    def vars(self, _):
+        raise TypeError('Unsupported assignment operation')
+
+    def get(self):
+        # Returns the synchronized hierarchical order graph.
+        if self._sync:
+            ds_state = self.resource.session.get(
```
jjdelc: You should almost never need to use `resource.session.get()` on this case (and all the above usages) are causing 2 GET requests to the same resource, the value you have under `self.resource.state` is what you want already, by wrapping it in `session.get()` you are making another exact GET to the same thing. Just do `ds_state = self.resource.state`

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+                self._load_graph()
+        return self
+
+
+class ProjectDatasetsOrder(Order):
+    def __init__(self, resource):
+        super(ProjectDatasetsOrder, self).__init__(resource)
+        self._datasets = None
+
+    def _load_hier(self):
```
jjdelc: Almost all the subclass methods here can disappear. Just have the subclass be like this:
```
def __init__(self, order, catalog):
    super()...
    self.datasets = catalog.by('id')
```

And you won't need any of the bottom resources.

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+            self._load_datasets()
+        return self._datasets
+
+    @datasets.setter
+    def datasets(self, _):
+        raise TypeError('Unsupported assignment operation')
+
+    def get(self):
+        # Returns the synchronized hierarchical order graph.
+        if self._sync:
```
jjdelc: Looks like you don't need this `_sync` method, since there is no need for lazy loading the resources, you already have them on init, the instance initialized by itself is good enough.

scrunch/order.py
```diff
@@ -0,0 +1,576 @@
...
+
+        def _get(group):
+            _elements = []
+            for obj in group.elements.values():
+                if isinstance(obj, Group):
+                    _elements.append({
+                        obj.name: _get(obj)
+                    })
+                else:
+                    if isinstance(obj, scrunch.datasets.Variable):
```
jjdelc: How does this work for the ProjectDatasetsOrder? The instance inside each will be a Dataset and not a Variable. 

Looks like part of this `_build_graph_structure` can be delegated to the subclasses where they know what each of their elements class are.

jjdelc:
👍

PR #158, opened 2017-08-09
added support to `integrate` derived variables

### LINKED ISSUE #150: Integration of derived variables
Add a way so we can ask Crunch to integrate (materialize) derived variables.

skriems:
closes #150


xbito:
I believe the docs ticket has been merged, which means that they have deployed a version that supports this operation. Merging this now.

Issue #165, opened 2017-09-05
Set values as missing for numeric variable


jamesrkg:
We need to be able to set a value as missing on a numeric variable. The API documentation is not clear on this. Under the PATCH section for variables it says that altering missing rules "are best or only achieved elsewhere", but it doesn't say where.

xbito:
Seems like we have done that manually in the past. It should be possible to add something to scrunch to tackle this.

jjdelc:
We have a `/missing_rules/` endpoint under each variable, a PUT request to add/edit/delete missing rules.
http://docs.crunch.io/#missing-values

mathiasbc:
`variable.set_missing_rules` allows this already.

PR #167, opened 2017-09-07
For #165 added ability to manage missing rules for a variable

dmonroy:
as described in #165, with this PR I'm adding the abilities to set and retrieve missing rules for a given variable.


dmonroy:
I MUST have said that I included 1 attribute and a method, let me describe it below:

- variable.missing_rules <= returns the dictionary object of the actual missing rules
- variable.set_missing_rules(rules) <= receives the dictionary of rules (reason as key, rule as value)

PR #202, opened 2017-11-13
Run tests in tox in Travis

jaraco:
For simplicity and to enable testing the full test suite, this change replaces the miniconda-based install with tox install in Travis. Some additional tweaks were necessary to get the tests passing, but they are now.


mathiasbc:
Should be ready to merge.

Issue #133, opened 2017-05-25
Filter name encoding error


jamesrkg:
An error is thrown if I try to create a filter with a non-ASCII character in its name.

```python
>>> ds.add_filter(expr='Z_geschlecht == 1', name='männlich')

KeyErrorTraceback (most recent call last)
<ipython-input-13-326f17f89b05> in <module>()
----> 1 ds.add_filter(expr='Z_geschlecht == 1', name='männlich')

C:\ProgramData\Anaconda2\lib\site-packages\scrunch\scrunch\datasets.pyc in add_filter(self, name, expr, public)
   1403                                  is_public=public))
   1404         new_filter = self.resource.filters.create(payload)
-> 1405         return self.filters[new_filter.body['name']]
   1406 
   1407     def create_single_response(self, categories, name, alias, description='',

KeyError: 'm\xc3\xa4nnlich'
```

xbito:
@jamesrkg we are going to need more context for this ticket. What python version are you running this on, which Operating System too?

jjdelc:
@jamesrkg are you using python 2 or 3?

skriems:
I talked to him and he uses Win10 and Py2.7

skriems:
@jamesrkg you can fix this in your scripts with either
```
from __future__ import unicode_literals
```
or you need to create your string literals like `u'männlich'`

skriems:
@jamesrkg if that doesn't solve your problem you're welcomed to reopen

Issue #134, opened 2017-05-25
Encoding error printing categories


jamesrkg:
An error is thrown if I try to print a categories object that contains non-ASCII characters:

```python
>>> print ds['Z_geschlecht'].categories

UnicodeEncodeErrorTraceback (most recent call last)
<ipython-input-16-e57a67be29b6> in <module>()
----> 1 print ds['Z_geschlecht'].categories

C:\ProgramData\Anaconda2\lib\collections.pyc in __repr__(self, _repr_running)
    188             if not self:
    189                 return '%s()' % (self.__class__.__name__,)
--> 190             return '%s(%r)' % (self.__class__.__name__, self.items())
    191         finally:
    192             del _repr_running[call_key]

UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 69: ordinal not in range(128)
```

skriems:
this one is still puzzling me and I'm inclined to blame the windows console/command prompt for this, which still tries to be compatible with systems which brought to the moon.

win10/cmd.exe
```
(' locale.getpreferredencoding()', '->', "'cp1252'")
('           sys.stdout.isatty()', '->', 'True')
('           sys.stdout.encoding', '->', "'cp65001'")
('            sys.stdin.isatty()', '->', 'True')
('            sys.stdin.encoding', '->', "'cp65001'")
('           sys.stderr.isatty()', '->', 'True')
('           sys.stderr.encoding', '->', "'cp65001'")
('      sys.getdefaultencoding()', '->', "'ascii'")
('   sys.getfilesystemencoding()', '->', "'mbcs'")
```
unix
```
 locale.getpreferredencoding() -> 'UTF-8'
           sys.stdout.isatty() -> True
           sys.stdout.encoding -> 'UTF-8'
            sys.stdin.isatty() -> True
            sys.stdin.encoding -> 'UTF-8'
           sys.stderr.isatty() -> True
           sys.stderr.encoding -> 'UTF-8'
      sys.getdefaultencoding() -> 'utf-8'
   sys.getfilesystemencoding() -> 'utf-8'
```
I came across https://github.com/Drekin/win-unicode-console and http://bugs.python.org/issue1602... I might be on a wrong track but anything I tried to get around this error failed.

I have have to further investigate when I have time

xbito:
@jamesrkg can you provide additional info of the context where this happens? OS, python version, language config info as Sebastian pasted?

xbito:
[...]

PR #223, opened 2017-12-27
Filter name encoding error

anderstheet:
Fix for:
https://github.com/Crunch-io/scrunch/issues/133

We don't get an encoding error anymore. But the naming is still not optimal.

Output look like this now:
{u'm\xe4nnlich': <Filter: name='m?nnlich'; id='5b1fb7db05d048998d353f3f9deda8d1'>, u'gender filter': <Filter: name='gender filter'; id='b08e5f73c8ab4e0ba4536d2515600ee6'>}

Obejct name can't be anything but ascii it seems in python 2:
https://bugs.python.org/issue5876#msg195996

### DIFF
--- a/scrunch/categories.py
+++ b/scrunch/categories.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+
 from six import PY2
 
 from scrunch.helpers import ReadOnly
--- a/scrunch/categories.py
+++ b/scrunch/categories.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from six import PY2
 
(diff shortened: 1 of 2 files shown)


anderstheet:
Made a fix for this ticket as well:
https://github.com/Crunch-io/scrunch/issues/134

Same as before. The nonascii chars are `?`s

xbito:
I don't feel 100% comfortable with the changes here, but at least it stops throwing an encoding error for past issues. @jamesrkg  feel free to review preview cases of encoding errors.

Issue #86, opened 2017-04-06
Subvariable iteration is random


jamesrkg:
`R6e` is a `multiple_response` array. I want to iterate over its `subvaraibles` using `for subvar in ds['R6e']:`, however they seem to be iterated in a random order.

I believe this is because `Variable` is ultimately inheriting from `collections.Mapping` via `DatasetVariablesMixin`.

We need the iteration over subvariables to be ordered.

```python
"alias": "R6e", 
"subvariables": [
	" ... /variables/000058/subvariables/0001/", 
	" ... /variables/000058/subvariables/0002/", 
	" ... /variables/000058/subvariables/0003/", 
	" ... /variables/000058/subvariables/0004/", 
	" ... /variables/000058/subvariables/0005/", 
	" ... /variables/000058/subvariables/0006/", 
	" ... /variables/000058/subvariables/0007/", 
	" ... /variables/000058/subvariables/0008/", 
	" ... /variables/000058/subvariables/0009/", 
	" ... /variables/000058/subvariables/0010/", 
	" ... /variables/000058/subvariables/0011/", 
	" ... /variables/000058/subvariables/0012/", 
	" ... /variables/000058/subvariables/0013/", 
	" ... /variables/000058/subvariables/0014/", 
	" ... /variables/000058/subvariables/0015/", 
	" ... /variables/000058/subvariables/0016/", 
	" ... /variables/000058/subvariables/0017/", 
	" ... /variables/000058/subvariables/0018/", 
	" ... /variables/000058/subvariables/0019/", 
	" ... /variables/000058/subvariables/0020/", 
	" ... /variables/000058/subvariables/0021/", 
	" ... /variables/000058/subvariables/0022/", 
	" ... /variables/000058/subvariables/0023/", 
	" ... /variables/000058/subvariables/0024/", 
	" ... /variables/000058/subvariables/0025/", 
	" ... /variables/000058/subvariables/0026/", 
	" ... /variables/000058/subvariables/0027/", 
	" ... /variables/000058/subvariables/0028/", 
	" ... /variables/000058/subvariables/0029/", 
	" ... /variables/000058/subvariables/0030/"
], 
```

```python
>>> for subvar in ds['R6e']:
>>>     print subvar.id
```

```
0012
0013
0003
0004
0023
0010
0001
0002
0022
0016
0011
0029
0028
0014
0006
0018
0017
0019
0007
0026
0015
0021
0027
0005
0020
0025
0009
0024
0008
0030
```

[...]

PR #229, opened 2018-01-18
Subvariable iteration is random

anderstheet:
Fixes: https://github.com/Crunch-io/scrunch/issues/86

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -529,7 +529,8 @@ class DatasetVariablesMixin(collections.Mapping):
         self.order = DatasetVariablesOrder(self._catalog, order)
 
     def __iter__(self):
-        for var in self._vars:
+        data = sorted(self._vars, key=lambda tup: tup[0])
+        for var in data:
             yield var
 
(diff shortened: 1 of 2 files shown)


xbito:
@anderstheet this introduces a sort to the subvariables, what should be happening is that we should be respecting the subvariable order established in the subvariables list of the variable definition that came from crunch. That is supposed to tell you the ordering. We can't trust the ordering from the subreferences dict because that is not guaranteed to be ordered.

scrunch/datasets.py
```diff
@@ -529,8 +529,8 @@ def _reload_variables(self):
         self.order = DatasetVariablesOrder(self._catalog, order)
 
     def __iter__(self):
-        for var in self._vars:
-            yield var
+        for subvar in self.resource['body']['subvariables']:
```
mathiasbc: this line will make scrunch fail, this method is used for handling Variables of a Dataset and Subvariables of a Variable, so in the case of Variables it wil fail with a KeyError. Dataset has no key subvariables.

PR #250, opened 2018-04-13
Folders 154061511

jjdelc:
Creates a new `BaseDataset.folders` attribute that allows to interact with the Folders API


jjdelc:
Implements a new `Dataset.folders` attribute that contains all the folder related operations. It is only available if such dataset has `variable_folders` flag enabled.

This change does not change the current `Order` Scrunch API leaving all existing scripts compatible with the current `/variables/hier/` endpoint. 

The methods:

 * `.get(path)`: Receives a Scrunch path string (| separated) and returns the last subfolder or variable indicated. Unless done of a specific folder it assumes the root is being used. Accessing an invalid path raises `InvalidPathError`
* `.folders.root`, `.folders.hidden`, `.folders.trash`: Returns a Folder instance of such folder from which you can .get() any subfolder or variable
* `Folder.make_subfolder()`, method in any Folder instance creates a subfolder under it. Receives only a name as parameter and return the newly created subfolder.
* `Folder.reorder()` receives an ordered list of Folder and Variable instances, will update the API `graph` attribute to the order of the received list.
* `Folder.move_here()`, takes a single instance, a list or *args of instances of either Folder and Variables and moves all those out of their current folder (wherever that is) inside the instance where this method is being called from.
* `Folder.rename()`, takes a string and renames the current folder
* `Folder.delete()`, performs a DELETE call on the current folder.

The branch includes integration tests that run with our CI

integration/test_folders.py
```diff
@@ -0,0 +1,371 @@
...
+        sf.move_here(self.ds['testvar4'], position=1)
+        self.assertEqual([c.name for c in sf.children], ['testvar2', 'testvar4', 'testvar3', 'testvar1'])
+
+    def test_move_by_alias(self):
+        root = self.ds.folders.root
+        target = root.create_folder("test_move_by_alias")
+        target.move_here('testvar2')
+        self.assertEqual([c.name for c in target.children], ["testvar2"])
+
+    def teat_make_folder_in_position(self):
```
bertjwregeer: s/teat/test/

This test has never been run...

setup.py
```diff
@@ -34,7 +34,7 @@
     python_requires='>=2.7',
     install_requires=[
         'pandas<0.20',
```
bertjwregeer: Is this a hard limit? And what is this hard limit for? I'd recommend loosening this pin a little if possible (with an environment marker if this project also needs to run on Python 3.4, let me know if this is the case and I can update this list of requires.)

PR #254, opened 2018-06-01
Use custom session class with custom user agent

jjdelc:
Make Scrunch report its own user agent and not Pycrunch's

### DIFF
--- a/scrunch/session.py
+++ b/scrunch/session.py
@@ -1,0 +1,15 @@
+from pycrunch import connect as _connect
+from pycrunch.elements import ElementSession
+
+from .version import __version__
+
+
+class ScrunchSession(ElementSession):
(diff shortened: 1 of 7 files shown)


scrunch/session.py
```diff
@@ -0,0 +1,15 @@
+from pycrunch import connect as _connect
+from pycrunch.elements import ElementSession
+
+from .version import __version__
+
+
+class ScrunchSession(ElementSession):
+    headers = {
+        "user-agent": "scrunch/%s" % __version__
```
nealrichardson: We also want to see the pycrunch version, so append this to it rather than overwrite it.

jjdelc: Added pycrunch's version as well

Issue #259, opened 2018-07-19
ds.folders.root.children only returns Folder instances


jamesrkg:
In the below, `item_url` is always in `index` and so `Folder` instances are returned for every item.

https://github.com/Crunch-io/scrunch/blob/f89f6541feaeb19e297978e3709df1cfe5ce1d0f/scrunch/folders.py#L129-L136

jjdelc:
Thanks for the report, here's the PR to fix it https://github.com/Crunch-io/scrunch/pull/260/files 
I'll be merging and releasing this.

PR #260, opened 2018-07-23
Return the right type of variables and folders

jjdelc:


scrunch/folders.py
```diff
@@ -127,12 +127,15 @@ def children(self):
...
+            if item_url not in index:
+                continue
+            item_tup = index[item_url]
+            if item_tup['type'] == 'folder':
+                subfolder = Folder(item_tup.entity, self.root, self)
                 _children.append(subfolder)
             else:
                 # Add the variable
-                _children.append(ds[item_url])
+                _children.append(ds[item_tup['alias']])
```
jamesrkg: Could we make this `_children.append(item_tup['alias'])` instead so that iteration is faster? Users can then access only the variables they need to manipulate later.

jjdelc: That would be too disruptive returning sometimes strings and sometimes Folder instances.

jjdelc:
Fixes issue #259

Issue #261, opened 2018-07-26
compare_dataset() - request for a new feature


GeorgianSorinDarie:
Hi,

Currently compare_dataset() method is returning the differences between two datasets, in terms of aliases, names, categories and data type. Would it be possible to also check the missing_rules() of each variable and  return the differences?

Thanks,
Giani

PR #269, opened 2018-08-07
compare_dataset() - request for a new feature

anderstheet:
fixing this issue:
https://github.com/Crunch-io/scrunch/issues/261

### DIFF
--- a/scrunch/mutable_dataset.py
+++ b/scrunch/mutable_dataset.py
@@ -206,7 +206,14 @@ class MutableDataset(BaseDataset):
             # 6. missing rules mismatch
             if self[name].type not in CATEGORICAL_TYPES and dataset[name].type not in CATEGORICAL_TYPES:
                 if self[name].missing_rules != dataset[name].missing_rules:
-                    diff['variables']['by_missing_rules'].append(name)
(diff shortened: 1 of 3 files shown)


scrunch/mutable_dataset.py
```diff
@@ -116,83 +116,15 @@ def join(self, left_var, right_ds, right_var, columns=None,
...
-
-                for sv_name in common_subnames:
-                    if self[name][sv_name].alias != dataset[name][sv_name].alias:
-                        if diff['subvariables'].get(name):
-                            diff['subvariables'][name].append(dataset[name][sv_name].alias)
-                        else:
-                            diff['subvariables'][name] = []
-                            diff['subvariables'][name].append(dataset[name][sv_name].alias)
-        return diff
+        url = '{}batches/compare/?dataset={}'.format(self.url, dataset.url)
```
jjdelc: if you have the Dataset() instance as a variable here, or the shoji resource then you should not be stitching URLs together, you should be able to do `dataset.resource.batches.follow('compare', 'dataset={}')` or something like that following the shoji navigation intead of hardcoding urls.

PR #279, opened 2018-08-16
adding edit_resolution method to support changing rollup_resolution of a datetime variable

mathiasbc:
This method is still not working, but I can't figure why. I am sending the data just like the UI is.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2157,7 +2157,7 @@ class Variable(ReadOnly, DatasetSubvariablesMixin):
     A pycrunch.shoji.Entity wrapper that provides variable-specific methods.
     DatasetSubvariablesMixin provides for subvariable interactions.
     """
-    _MUTABLE_ATTRIBUTES = {'name', 'description',
+    _MUTABLE_ATTRIBUTES = {'name', 'description', 'rollup_resolution',
                            'view', 'notes', 'format', 'derived'}
     _IMMUTABLE_ATTRIBUTES = {'id', 'alias', 'type', 'discarded'}
     # We won't expose owner and private
@@ -2502,7 +2502,5 @@ class Variable(ReadOnly, DatasetSubvariablesMixin):
         assert self.type == 'datetime', 'Method only allowed for datetime variables'
         self.dataset._validate_vartypes(self.type, resolution=resolution)
 
-        payload = json.dumps({'rollup_resolution': resolution})
-        resp = self.resource.patch(payload)
-        self.dataset._reload_variables()
-        return self.dataset[self.alias]
+        self.resource.edit(rollup_resolution=resolution)
+        return self
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -41,7 +41,7 @@ LOG = logging.getLogger('scrunch')
 CATEGORICAL_TYPES = {
     'categorical', 'multiple_response', 'categorical_array',
 }
-
+RESOLUTION_TYPES = ['Y', 'Q', 'M', 'W', 'D', 'h', 'm', 's', 'ms']
 
 def _set_debug_log():
     # ref: http://docs.python-requests.org/en/master/api/#api-changes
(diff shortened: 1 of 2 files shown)


mathiasbc:
@jjdelc This is failing with:
```
pycrunch.lemonpy.ClientError: (<Response [400]>, 'https://app.crunch.io/api/datasets/b8446edded684f03a4388326266b80ff/variables/1c95095169314754924ccdf6ca879d8b/', {'status': '400 Bad Request', 'message': 'Invalid variable attributes: rollup_resolution'})
```
Strange, from my understanding the UI is doing a PATCH to the variable endpoint with a payload of `{'resolution_rollup': resolution}`

jjdelc:
The attribute is `resolution`, send it as `{resolution: new_resolution}`

scrunch/datasets.py
```diff
@@ -2495,3 +2493,16 @@ def transformations(self):
...
+        """
+        PATCHes the rollup_resolution attribute of a datetime variable. This is the
+        equivalent to the UI's change resolution action.
+        """
+        assert self.type == 'datetime', 'Method only allowed for datetime variables'
+        self.dataset._validate_vartypes(self.type, resolution=resolution)
+
+        payload = json.dumps({'rollup_resolution': resolution})
+        resp = self.resource.patch(payload)
+        self.dataset._reload_variables()
```
jjdelc: If you have `self.resource` to be the Variable shoji entity, you only need to do `self.resource.edit(resolution=new_resolution)` no need to do json.dumps.

scrunch/datasets.py
```diff
@@ -2495,3 +2493,16 @@ def transformations(self):
...
+        return self.dataset[self.alias]
```
jjdelc: If you use the `.edit()` pycrunch method, then the resource's attribute will get automatically updated so no need to re-read the Variable instance or to do a full reload of all the variables (can be a expensive GET). You could just return `self`.

mathiasbc:
@jjdelc: resolution is a different attribute and in this case we don't intend to change that. @jamesrkg noticed that the UI only changes `resolution_rollup ` and `resolution` remains the same. The payload sent by the UI is exactly as I pasted above.

scrunch/datasets.py
mathiasbc: yes that the original intention but since it was failing I decided to test other paths. Neither of the methods works for this attribute.

scrunch/datasets.py
mathiasbc: 👍

jjdelc:
Ah, I see. The attribute is `rollup_resolution`, "rollup" is first. http://docs.crunch.io/endpoint-reference/endpoint-variable.html

mathiasbc:
@jjdelc: Ah sorry that was my mistake in the comment, but the code actually had the correct name. I made some updates, still I get:
pycrunch.lemonpy.ClientError: (<Response [400]>, 'https://app.crunch.io/api/datasets/b8446edded684f03a4388326266b80ff/variables/1c95095169314754924ccdf6ca879d8b/', {'status': '400 Bad Request', 'message': 'Invalid variable attributes: rollup_resolution'})

mathiasbc:
@jjdelc @jamesrkg @xbito This should be good to go now. Changed a few things. Fernando helped me realized we need to change the value inside the `view` attribute and not the external `rollup_resolution`. Renamed the method to `edit_resolution` to avoid the confusion with the variable attribute `rollup_resolution`. 

Note: for this method to work the user needs to grab the Dataset editor's lock. `get_dataset(ds_id, editor=True)`

scrunch/datasets.py
```diff
@@ -2495,3 +2493,18 @@ def transformations(self):
...
+        equivalent to the UI's change resolution action. 
+
+        Be sure to grab the editor's lock of the dataset.
+        
+        :usage: edited_var = var.edit_resolution('M')
+        assert editar_var.rollup_resolution == 'M'
+        """
+        assert self.type == 'datetime', 'Method only allowed for datetime variables'
+        self.dataset._validate_vartypes(self.type, resolution=resolution)
+        self.resource.edit(view=dict(rollup_resolution=resolution))
```
jjdelc: This is wrong, note that here you will overwrite all the other attributes in the variable's view attribute. You need to read the current dict, then change the value an send that as the new dict. Sending a new dict will overwrite the full thing.

PR #283, opened 2018-08-20
closes 282

mathiasbc:


scrunch/tests/test_datasets.py
```diff
@@ -1218,10 +1218,26 @@ def test_edit_alias(self):
...
 
+    def test_edit_resolution(self):
+        variables = {
+            '001': {
+                'id': '001',
+                'alias': 'datetime_var',
+                'name': 'Datetime Variable',
+                'type': 'datetime',
+                'is_subvar': False,
+                'view': {'rollup_resolution': 'ms'}
```
jjdelc: In order to make the test verify that it's not overwriting other attributes you need to have this initial `view` value contain some other key. And then assert that both keys are indeed still there.

scrunch/tests/test_datasets.py
```diff
@@ -1218,10 +1218,26 @@ def test_edit_alias(self):
...
+            },
+        }
+        ds_mock = self._dataset_mock(variables=variables)
+        ds = MutableDataset(ds_mock)
+        ds.resource = mock.MagicMock()
+        var = ds['datetime_var']
+        var.edit_resolution('M')
+        var.resource._edit.assert_called_with(view={'rollup_resolution': 'M'})
```
jjdelc: This assertion would pass even with the original code before this fix landed. Needs to have another attribute.

scrunch/tests/test_datasets.py
jjdelc: In order to make the test verify that it's not overwriting other attributes you need to have this initial view value contain some other key. And then assert that both keys are indeed still there.

scrunch/tests/test_datasets.py
jjdelc: In order to make the test verify that it's not overwriting other attributes you need to have this initial view value contain some other key. And then assert that both keys are indeed still there.

PR #285, opened 2018-08-21
Feature/267 crunch teams

mathiasbc:
This is a non-backward compatible change, since it will be removing methods from the Project class and moving them to a generic approach `add_member`, `remove_member`, `members` that will work equally with Teams or Projects.

This is still a work in progress because it will need tests and I am seeking design approval first.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -212,6 +212,25 @@ def get_user(user, connection=None):
     return User(ret)
 
 
+def get_team(team, connection=None):
+    """
+    :param team: Crunch Team Name (crunch only lists teams by Name)
+    :param connection: An scrunch session object
+    :return: Team class instance
+    """
+    if connection is None:
+        connection = _get_connection()
+        if not connection:
+            raise AttributeError(
+                "Authenticate first with scrunch.connect() or by providing "
+                "config/environment variables")
+    try:
+        ret = connection.teams.by('name')[team].entity
+    except KeyError:
+        raise KeyError("Team name: %s not found." % team)
+    return Team(ret)
+
+
 def list_geodata(name=None, connection=None):
     """
     :param connection: An scrunch session object
@@ -267,22 +286,37 @@ class Team:
     _IMMUTABLE_ATTRIBUTES = {'id'}
     _ENTITY_ATTRIBUTES = _MUTABLE_ATTRIBUTES | _IMMUTABLE_ATTRIBUTES
 
-    def __init__(self, resource):
-        self.resource = resource
+    def __init__(self, team_resource):
+        self.resource = team_resource
         self.url = self.resource.self
 
     def __getattr__(self, item):
         if item in self._ENTITY_ATTRIBUTES:
             return self.resource.body[item]
-
-        # Attribute doesn't exists, must raise an AttributeError
         raise AttributeError('Team has no attribute %s' % item)
 
     def __repr__(self):
-        return "<Team: name='{}' id='{}'>".format(self.name, self.id)
+        return "<Team: name='{}'; id='{}'>".format(self.name, self.id)
 
     def __str__(self):
         return self.name
+
+    @property
+    def users(self):
+        """
+        :return: dictionary of User instances
+        """
+        # TODO: return a dictionary keyed by email and values should be User
+        # instances, but when trying got 403 from Crunch and we need
+        # support for teams now
+        users = []
+        for member in self.resource.members.index.values():
+            # members can be users or teams
+            user = member.get('email')
+            if not user:
+                user = member.get('name')
+            users.append(user)
+        return users
 
 
 class Project:
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -26,7 +26,9 @@ from scrunch.subentity import Deck, Filter, Multitable
 from scrunch.variables import (combinations_from_map, combine_categories_expr,
                                combine_responses_expr, responses_from_map)
 
+
 _MR_TYPE = 'multiple_response'
+
 
 if six.PY2:  # pragma: no cover
     from urlparse import urljoin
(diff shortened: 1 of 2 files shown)


jjdelc:
If we're going forward for this change, I'd suggest using a helper class instead of a mixin. It will help us keep the Project's class interface less loaded.

```
class Members:
  def __init__(self, resource):
      self.resource = resource

  def add(self, member):...
  def remove(self, member):...
  def edit(self, member):...
  def list(self):...

class Project:
  @property
   def members(self):
      return Members(self.resource)


>>> project.members.add(new_member)
>>> project.members.list(new_member)
```

Opinions about this approach? I think that in Scrunch we suffer from classes having too many methods and encapsulate them like this could help a bit.

xbito:
I agree with Jj suggestion. On top of that though, we don't make backwards incompatible changes, not without a couple of versions (months in time probably).
So, we must provide a compatibility layer, which I believe should be pretty straightforward. Do a Python Warning when using the old methods, point to a wiki page with backwards incompatible changes that are coming and suggesting a migration path.

mathiasbc:
@jjdelc: ^

mathiasbc:
@jjdelc: now that this is merged with master there are tests failing because of the addition of `test_projects` file. These tests are failing for master as well.

jjdelc:
Are you on latest master? I released 0.6.3 on Thursday after merging the test_projects file in and got travis to successfully pass. What errors are you getting?

mathiasbc:
@jjdelc: yeah I pulled master, this is the output from `master`:

```
Results (6.43s):
     248 passed
       8 failed
         - scrunch/tests/test_projects.py:49 TestProjectNesting.test_create_subproject
         - scrunch/tests/test_projects.py:180 TestProjectNesting.test_follow_path
         - scrunch/tests/test_projects.py:359 TestProjectNesting.test_is_root
         - scrunch/tests/test_projects.py:342 TestProjectNesting.test_move
         - scrunch/tests/test_projects.py:209 TestProjectNesting.test_move_things
         - scrunch/tests/test_projects.py:269 TestProjectNesting.test_place
         - scrunch/tests/test_projects.py:195 TestProjectNesting.test_rename
         - scrunch/tests/test_projects.py:325 TestProjectNesting.test_reorder
       3 xfailed
 mbc@MacBook-Pro  ~/Yougov/G/scrunch   master 
```

mathiasbc:
@jjdelc: A missing dependency apparently.

scrunch/datasets.py
```diff
@@ -262,28 +295,96 @@ def __str__(self):
...
+        :return: A list of members of the Entity
+        """
+        members = []
+        for member in self.resource.members.index.values():
+            # members can be users or teams
+            user = member.get('email')
+            if not user:
+                user = member.get('name')
+            members.append(user)
+        return members
```
jjdelc: Is this returning a list of strings? Include in the docstring what is the expected result. 

If this returns only a list of emails, I think it'd be better to name the function `emails` or something like that.

scrunch/datasets.py
```diff
@@ -262,28 +295,96 @@ def __str__(self):
...
+        self.resource = resource
+
+    def list(self):
+        """
+        :return: A list of members of the Entity
+        """
+        members = []
+        for member in self.resource.members.index.values():
+            # members can be users or teams
+            user = member.get('email')
```
jjdelc: On which case there is not an email available?

scrunch/datasets.py
```diff
@@ -262,28 +295,96 @@ def __str__(self):
...
+        """
+        member = self._validate_member(member)
+        self.resource.members.patch({member.url: None})
+
+    def add(self, member, edit=False):
+        """
+        :param member: email, User instance, team name or Team instance
+        :return: None
+        """
+        member = self._validate_member(member)
```
jjdelc: Note that the API allows to add members directly by email and if they don't exist they will be created as new user accounts on Crunch. Doing `_validate_member` here removes that option to "invite" new members by email directly since this would raise for a non existing new user.

scrunch/datasets.py
```diff
@@ -319,14 +420,30 @@ def __repr__(self):
...
+    @property
+    def members(self):
+        return Members(self.resource)
+
+    def get_dataset(self, dataset):
+        try:
+            shoji_ds = self.resource.datasets.by('name')[dataset].entity
+        except KeyError:
+            try:
+                shoji_ds = self.resource.datasets.by('id')[dataset].entity
```
jjdelc: Factor out `self.resource.datasets` here since you are doing that GET request two times.

scrunch/datasets.py
```diff
@@ -319,14 +420,30 @@ def __repr__(self):
...
         """
-        :return: dictionary of User instances
+        TODO: deprecate in favor of members.list property
         """
-        # TODO: return a dictionary keyed by email and values should be User
-        # instances, but when trying got 403 from Crunch and we need
-        # support for teams now
+        LOG.warning("""This method is legacy and will be deprecated
+            in future releases. Please make use of project.members.list()
+            instead""")  # noqa: E501
```
jjdelc: Why not rewrite the body of these functions to just use the new code, then they're only shortcuts and not duplicate of logic.  Same for the other deprecated methods.

scrunch/datasets.py
mathiasbc: if the member it's a Team

scrunch/datasets.py
mathiasbc: Look how this method is designed, since the beginnings of time. We are patching the `/members/` resource with `self.resource.members.patch({member.url: {'edit': edit}})` So it needs to be an instance with a url. I'm happy to explore other options, but I believe this is to be addressed on a different independent ticket.

scrunch/datasets.py
mathiasbc: good idea.

scrunch/datasets.py
jjdelc: Great then, if it's been like that we could add another method later to invite.

PR #290, opened 2018-09-04
include derive_multiple_response method to create more flexible categ…

mathiasbc:
…orial array

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1230,7 +1230,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         :param description: Description of the variable to create
         :param notes: Notes of the variable to create
 
-        Note: It's important that each subvariable defines the same numbes of cases for every category.
+        Note: It's important that each subvariable defines the same number of cases for every category.
         Suvariable alias will be derived from the alias to the variable + subvariable id to keep
         compliance with other methods in Scrunch
         """
@@ -1241,7 +1241,6 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
             _validate_category_rules(categories, subvar['cases'])
 
         responses_map = collections.OrderedDict()
-        responses_map_ids = []
         for subvar in subvariables:
             _cases = []
             for case in subvar['cases'].values():
@@ -1250,7 +1249,6 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                     _cases.append(_case)
 
             resp_id = '%04d' % subvar['id']
-            responses_map_ids.append(resp_id)
             responses_map[resp_id] = case_expr(
                 _cases,
                 name=subvar['name'],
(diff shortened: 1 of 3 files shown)


scrunch/datasets.py
```diff
@@ -1207,11 +1207,84 @@ def rollup(self, variable_alias, name, alias, resolution, description='',
...
+        compliance with other methods in Scrunch
+        """
+        # validate categories and fill defaults
+        categories = validate_categories(categories)
+        # validate that every subvar defines rules for all categories
+        for subvar in subvariables:
+            _validate_category_rules(categories, subvar['cases'])
+
+        responses_map = collections.OrderedDict()
+        responses_map_ids = []
```
jjdelc: Why do you need this separate list? Could you just use `responses_map.keys()`, since it's an ordered dict, its keys will also be ordered.

scrunch/helpers.py
```diff
@@ -7,6 +7,12 @@
     from urllib.parse import urljoin
 
 
+DEFAULT_CATEGORIES = [
```
jjdelc: Maybe rename this to `DEFAULT_MULTIPLE_RESONSE_CATEGORIES` or something to make it a bit more clear that this is for MR vars

scrunch/helpers.py
```diff
@@ -70,7 +76,50 @@ def download_file(url, filename):
...
+        if category.get('missing'):
+            missing_count += 1
+        if not category.get('id'):
+            raise ValueError('An "id" must be provided to all categories')
+        if not category.get('name'):
+            raise ValueError('A "name" must be provided to all categories')
+    if selected_count > 1 or selected_count == 0:
+        raise ValueError('Categories must define one category as selected')
+    if missing_count > 1:
+        raise ValueError('Categories can\'t have more than 1 category as missing')
```
jjdelc: Is this so? There is no restriction on the backend for this rule. You can have all the missing categories you want.

scrunch/helpers.py
```diff
@@ -70,7 +76,50 @@ def download_file(url, filename):
...
+    """
+    defaults = {'missing': False, 'numeric_value': None, 'selected': False}
+    selected_count = 0
+    missing_count = 0
+    for category in categories:
+        if category.get('selected'):
+            selected_count += 1
+        if category.get('missing'):
+            missing_count += 1
+        if not category.get('id'):
```
jjdelc: A few extra validations here:
 * ID must be an int
 * IDs must be unique

Similar for names:
  * Must be non empty strings
  * Must be unique
It's also OK if you don't do them though, the server will complain and won't let you anyway.

mathiasbc: This is cool, but should apply to all other methods that deal with variables, so it's perhaps a good idea to have a ticket to enforce this globally

mathiasbc:
@jjdelc changes made, the failing test seems to be a random timeout on one pything version

PR #293, opened 2018-09-16
derive_multiple_response in now callable from create_categorical in o…

mathiasbc:
…rder to create 3 categories mutliple response variables in a default setup

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1391,7 +1391,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         return self[alias]
 
     def create_categorical(self, categories, alias, name, multiple, description='',
-        notes='', missing=None):
+        notes='', missing_case=None):
         """
         Used to create new categorical variables using Crunchs's `case`
         function
@@ -1450,7 +1450,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                     }],
                 multiple=True
                 (B) If the missing_case is constant across all subvariables, then the argument
-                    `missing` can be passed as argument to this function:
+                    `missing_case` can be passed as argument to this function:
                 categories: [
                     {
                         'case': 'var_1 == 1',
@@ -1468,12 +1468,17 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                         'id': 3,
                     }],
                 multiple=True,
-                missing='missing(var_1)'
+                missing_case='missing(var_1)'
         """
-        # First we append the missing case to every subvariable and let the 
+        # Initially validate that we dont have `missing_case` argument and `missing_case`
+        # in the categories list
+        if any(['missing_case' in c.keys() for c in categories]) and missing_case:
+            raise ValueError('missing_case as an argument and as element of \
+                "categories" is not allowed.')
+        # First we append the missing_case to every subvariable and let the 
         # generic case deal with it
-        if missing:
-            default = {'missing_case': missing}
+        if missing_case:
+            default = {'missing_case': missing_case}
             _categories = []
             for sv in categories:
                 _default = default.copy()
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1428,7 +1428,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
 
             Then we need to declare the extra Missing case, which can be done in one of the
             follwing 2 ways:
-                (A) Every subvariable declares it's own missing_case individually in the 
+                (A) Every/some subvariable declare it's own missing_case individually in the 
                     `missing_case` element
                 categories: [
                     {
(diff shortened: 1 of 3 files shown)


scrunch/datasets.py
```diff
@@ -1398,7 +1398,50 @@ def create_categorical(self, categories, alias, name, multiple, description='',
...
+            _categories = [
+                {'id': 1, 'name': 'Selected', 'selected': True},
+                {'id': 2, 'name': 'Not Selected'},
+                {'id': 3, 'name': 'No Data', 'missing': True}
+            ]
+            _subvariables = []
+            for sv in categories:
+                _subvariables.append(
+                    {'id': sv['id'], 'name': sv['name'], 'cases': {1: sv['case'], 2: sv['missing_case']}}
+                )
```
jamesrkg: This is currently mapping `missing_case` to `not selected` but it should map to `No Data`.

mathiasbc: yes, this is a bug.

xbito:
👍

mathiasbc:
this one is done after some iterations.

scrunch/datasets.py
```diff
@@ -1391,15 +1391,139 @@ def create_numeric(self, alias, name, derivation, description='', notes=''):
...
+        """
+        # Initially validate that we dont have `missing_case` argument and `missing_case`
+        # in the categories list
+        if any(['missing_case' in c.keys() for c in categories]) and missing_case:
+            raise ValueError('missing_case as an argument and as element of \
+                "categories" is not allowed.')
+        # First we append the missing_case to every subvariable and let the 
+        # generic case deal with it
+        if missing_case:
+            default = {'missing_case': missing_case}
```
jjdelc: It looks like what you're doing here is adding the missing case to each of the received categories, I think you can simplify all this with:
````
for cat in categories:
    cat['missing_case'] = missing_case
```

scrunch/datasets.py
```diff
@@ -1391,15 +1391,139 @@ def create_numeric(self, alias, name, derivation, description='', notes=''):
...
+            _categories = []
+            for sv in categories:
+                _default = default.copy()
+                _default.update(sv)
+                _categories.append(_default)
+            categories = _categories
+
+        # In the case of MR and all cases declare a 'missing_case'
+        if multiple and any(['missing_case' in c.keys() for c in categories]):
```
jjdelc: You already calculated this above on the first validation? why not store it in a variable `cats_have_missing = any(...)`

scrunch/datasets.py
```diff
@@ -1391,15 +1391,139 @@ def create_numeric(self, alias, name, derivation, description='', notes=''):
...
+            ]
+            _subvariables = []
+            for sv in categories:
+                if 'missing_case' in sv:
+                    _subvariables.append(
+                        # Make all cases mutually exclusive by negations
+                        {
+                            'id': sv['id'],
+                            'name': sv['name'],
+                            'cases': {
```
jjdelc: It looks like it's only the `cases` key that's different between both conditions. I'd suggest to build a `case` variable with the `{id, name}` keys and only add the particular `cases` key to it inside the conditional. Then just do the append to `_subvariables` outside of the `if`.

scrunch/datasets.py
```diff
@@ -1391,15 +1391,139 @@ def create_numeric(self, alias, name, derivation, description='', notes=''):
...
+            for sv in categories:
+                _default = default.copy()
+                _default.update(sv)
+                _categories.append(_default)
+            categories = _categories
+
+        # In the case of MR and all cases declare a 'missing_case'
+        if multiple and any(['missing_case' in c.keys() for c in categories]):
+            _categories = [
+                {'id': 1, 'name': 'Selected', 'selected': True},
```
jjdelc: Looks like these IDs have semantics, do you think they could be stored as constants?

 * `SELECTED_ID = 1`
 * `NOT_SELECTED_ID = 2`
 * `NO_DATA_ID = 3`

And reuse those down below when constructing the cases.

scrunch/tests/test_datasets.py
```diff
@@ -1986,6 +1986,178 @@ def test_recode_multiple_response(self):
             }
         })
 
+    def test_create_categorical_missing_case(self):
```
jjdelc: I see you only have one test and you have added many conditions, looks like there's untested paths still? Let me know if they're already tested elsewhere in case it isn't necessary to add more scenarios.

scrunch/tests/test_datasets.py
```diff
@@ -1986,6 +1986,178 @@ def test_recode_multiple_response(self):
...
+                    'name': 'Twitter',
+                    'case': 'var_b == 1',
+                }
+            ]
+        }
+        with pytest.raises(ValueError) as err:
+            ds.create_categorical(**kwargs)
+        assert 'Entity test_dataset_name has no (sub)variable' in str(err.value) 
+        ds.resource.variables.create.assert_called_with({  
+            'element':'shoji:entity',
```
jjdelc: You could probably skip this boilerplate in your test using the shoji wrapper functions

PR #294, opened 2018-09-24
Force set to False the projects feature flag so tests use the new API

jjdelc:


integration/test_projects.py
```diff
@@ -14,7 +14,8 @@
 password = os.environ['SCRUNCH_PASS']
 
 site = connect(username, password, HOST)
-UNIQUE_PREFIX = unicode(datetime.now()).replace(':', '').replace('.', '')
+UNIQUE_PREFIX = str(datetime.now()).replace(':', '').replace('.', '')
```
jjdelc: Added this switch from `unicode` to `str` because python 3.5+ doesn't support it and Scrunch's codebase is python 3.x compatible. It wasn't a problem because we always run the integration tests with python 2.x, but I wanted to get rid of the IDE warning.

PR #299, opened 2018-10-10
pandas must be an optional requirement 

dmonroy:
pandas must be an optional requirement as no all applications that uses scrunch needs pandas.

Also pandas requires numpy and having both wastes unnecessary resources when building an environment from scratch.

As soon as this is merged I'll add the corresponding instructions to the wiki for the people needing pandas.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -6,7 +6,13 @@ import os
 import re
 import sys
 
-import pandas as pd
+try:
+    import pandas as pd
+except ImportError:
+    # pandas has not been installed, don't worry!
+    # ... unless you have to worry about pandas
+    pd = None
+
 import six
 
 import pycrunch
@@ -2013,6 +2019,8 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
             A DataFrame representation of all attributes from all forks
             on the given dataset.
         """
+        if pd is None:
+            raise ModuleNotFoundError("No module named 'pandas'")
 
         if len(self.resource.forks.index) == 0:
             return None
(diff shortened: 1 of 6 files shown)


scrunch/datasets.py
```diff
@@ -2013,6 +2019,8 @@ def forks_dataframe(self):
             A DataFrame representation of all attributes from all forks
             on the given dataset.
         """
+        if pd is None:
+            raise ModuleNotFoundError("No module named 'pandas'")
```
jjdelc: This message should read more like "Pandas is not installed, please install it in your environment to use this function" or something along those lines.

setup.py
```diff
@@ -60,7 +59,7 @@
 
             # local
         ],
-        'pandas': ['pandas'],
+        'pandas': ['pandas<0.20'],
```
jjdelc: Why the lock on this version?

scrunch/datasets.py
dmonroy: I was thinking on replicate the error that python raises when you import something that is not installed, but your message makes more sense

setup.py
dmonroy: when we remove that version constraint tests fails with: ```E           AttributeError: 'DataFrame' object has no attribute 'sort'```

dmonroy: and fixed it with `_forks.sort_values(by=['creation_time'], inplace=True)`

jjdelc:
Can you add a test verifying this behavior?

dmonroy:
@jjdelc what do you think now about this PR? 

pandas (and numpy) adds +200MB to our builds and lot of processing time to the pipeline

PR #300, opened 2018-10-10
Test moving datasets

jjdelc:
This branch adds an integration test to run on Crunch CI to replicate a couple of bugs.

Also updates the unit tests for this behavior.

### DIFF
--- a/integration/test_projects.py
+++ b/integration/test_projects.py
@@ -123,3 +123,4 @@ class TestProjects(TestCase):
         pb.resource.refresh()
         self.assertItemsEqual(pb.resource.index.keys(),
             [_ds1.self, _ds2.self, _ds4.self])
+        self.assertEqual(ds2.resource.project.self, pb.url)
--- a/integration/test_projects.py
+++ b/integration/test_projects.py
@@ -4,14 +4,16 @@ import os
 from datetime import datetime
 from unittest import TestCase
 
-from scrunch import connect, get_project
 from scrunch.datasets import Project
 from scrunch.helpers import shoji_entity_wrapper
+from scrunch import connect, get_project, get_dataset, get_user
 
 
 HOST = os.environ['SCRUNCH_HOST']
 username = os.environ['SCRUNCH_USER']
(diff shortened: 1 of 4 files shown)


integration/test_projects.py
```diff
@@ -71,3 +73,54 @@ def test_reorder(self):
         pa.reorder(["2", "1"])
         pa.resource.refresh()
         self.assertEqual(pa.resource.graph, [p2.url, p1.url])
+
+    def test_move_dataset(self):
```
jjdelc: This test runs against a test version of the Crunch API in our CI

scrunch/datasets.py
```diff
@@ -544,6 +544,8 @@ def move_here(self, items, **kwargs):
             item.url: {} for item in items
         }, graph=graph))
         self.resource.refresh()
+        for item in items:
```
jjdelc: This is the scrunch bug fix. After moving the items they should reflect their new location. Without this you'd move a dataset and then do `.project` and it'd be pointing to the old location.

scrunch/tests/test_projects.py
```diff
@@ -47,35 +47,44 @@ def test_detect_correct_handler(self):
         self.assertTrue(isinstance(project.order, Project))
 
     def test_create_subproject(self):
-        session = Mock(
-            feature_flags={'old_projects_order': False}
-        )
+        session = MockSession()
```
jjdelc: Replaced this test to use `MockSession` isntead of Mock

scrunch/tests/mock_session.py
```diff
@@ -16,19 +16,27 @@ class FixtureAdapter(BaseAdapter):
     def __init__(self):
         self.fixtures = {}
         self.requests = []
+        self.post_responses = []
 
     def add_fixture(self, url, fixture):
         self.fixtures[url] = fixture
 
+    def add_post_response(self, response):
```
jjdelc: `MockSession` didn't support POST requests, so had to add a few methods for the `create_subproject` test.

PR #301, opened 2018-10-11
Expose the dataset's size object as a property of the dataset instance

dmonroy:
So we (as scrunch users) are able to do:
```python
>>> dataset.size.columns
... 99
```

instead of:
```python
>>> dataset.resource.body.size.columns
... 99
```

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2514,6 +2514,13 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         }
         self.resource.table.post(json.dumps(payload))
 
+    @property
+    def size(self):
+        """
+        Exposes the dataset's size object as a property of the dataset instance
+        """
+        return self.resource.body.size
+
 
 class Dataset(BaseDataset):
 


jjdelc:
And a test so it doesn't get deleted/changed in the future?

scrunch/tests/test_datasets.py
```diff
@@ -439,6 +445,14 @@ def _session_get(*args):
         ds.create_crunchbox()
         ds_mock.boxdata.create.assert_called_with(expected_payload)
 
+    def test_dataset_size(self):
+        """
+        Using `ds.size` actually returns the value of `ds.resource.body.size`
+        """
+        resource = mock.MagicMock()
+        ds = StreamingDataset(resource)
+        assert str(ds.size.rows).startswith("<MagicMock name='mock.body.size.rows' id='")
```
jjdelc: Best to load the `resource` mock to have the right attributes (`Mock(size={'rows': 666})`) then you can assert that the size is 666, better than asserting over the Mock's instance `repr`

PR #302, opened 2018-10-11
Warsaw demo

slobodan-ilic:
- Implement `crtabs` - a method for seamless fetching of crunch cubes
- Add `cr.cube` dependency
- Needs [this pycrunch PR](https://github.com/Crunch-io/pycrunch/pull/54) to work properly
- Needs [this cr.cube PR](https://github.com/Crunch-io/crunch-cube/pull/105) to display cubes in console nicely

### DIFF
--- a/scrunch/cubes.py
+++ b/scrunch/cubes.py
@@ -1,0 +1,12 @@
+from pycrunch.cubes import fetch_cube, count
+from cr.cube.crunch_cube import CrunchCube
+
+
+def crtabs(dataset, variables):
+    """Return CrunchCube representation of crosstab.
+
+    :param dataset: Dataset shoji object
+    :param variables: List of variable urls, names or aliases
+    """
+    dataset = dataset.resource
+    return CrunchCube(fetch_cube(dataset, variables, count=count()))
--- a/scrunch/cubes.py
+++ b/scrunch/cubes.py
@@ -1,4 +1,5 @@
 from pycrunch.cubes import fetch_cube, count
+from scrunch.datasets import Variable
 from cr.cube.crunch_cube import CrunchCube
 
 
@@ -9,4 +10,12 @@ def crtabs(dataset, variables):
     :param variables: List of variable urls, names or aliases
     """
     dataset = dataset.resource
+    variables = prepare_variables(variables)
     return CrunchCube(fetch_cube(dataset, variables, count=count()))
+
+
+def prepare_variables(variables):
+    return [
+        v.url if type(v) is Variable else v
+        for v in variables
+    ]
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,5 @@ crunch.ini
 # Pycharm/Idea
 .idea/
 
+# Vim
+*.swp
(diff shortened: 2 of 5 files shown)


mathiasbc:
@slobodan-ilic: Something related https://github.com/Crunch-io/scrunch/blob/master/scrunch/subentity.py#L238. Which I think we might want to get rid of.

slobodan-ilic:
> @slobodan-ilic: Something related https://github.com/Crunch-io/scrunch/blob/master/scrunch/subentity.py#L238. Which I think we might want to get rid of.

Definitely! I saw that we're not using it in the codebase of `scrunch`. But it's an open-source lib after all, so there's no way to know if anyone's actually using it in their scripts. Can you confirm that it's not useful to you/your team/general pool of scrunch users/to the best of your knowledge?

jjdelc:
@slobodan-ilic  can you add a small test for this? Just asserting that at least CrunchCube gets called with the right parameters.

scrunch/cubes.py
```diff
@@ -0,0 +1,12 @@
+from pycrunch.cubes import fetch_cube, count
+from cr.cube.crunch_cube import CrunchCube
+
+
+def crtabs(dataset, variables):
+    """Return CrunchCube representation of crosstab.
+
+    :param dataset: Dataset shoji object
+    :param variables: List of variable urls, names or aliases
```
jjdelc: Note about this `variables`, chances are high this will be a `scrunch.Variable`, I'd recommend to also expect that and if they're such instances, convert them to that.

If you can add a test would help.

slobodan-ilic:
@jjdelc @mathiasbc I've added capability to handle `Variable` entries. Also supported with unit tests (under the assumption that we trust `fetch_cube` from pycrunch). I think this can go now...

jjdelc:
Cool, thanks

jjdelc:
I guess my only question now is if `crtabs` is the agreed name for the function? why not just `tabs`? 

`from scrunch.cubes import crtabs`

I am not sure if the `cr` prefix makes sense insude Scrunch, it's all the Crunch API anyway. @jamesrkg  do you have a more natural name for this function? I would propose `tabulate()`

slobodan-ilic:
> I guess my only question now is if `crtabs` is the agreed name for the function? why not just `tabs`?
> 
> `from scrunch.cubes import crtabs`
> 
> I am not sure if the `cr` prefix makes sense insude Scrunch, it's all the Crunch API anyway. @jamesrkg do you have a more natural name for this function? I would propose `tabulate()`

@jjdelc Actually the `crtabs` comes from the _wonderful_ world of `R`. And the shorthand `cr` in that phrase doesn't stand for Crunch (imagine that!!!). It's a shorthand for "crosstabs", which is a common term in statistics.

Anyways, I'm all in favor of whatever makes sense to the people who'll be using it the most (while still preserving common sense). So, please do share your opinion, and I'll take care of it @jamesrkg , @mathiasbc

jjdelc:
ohhh! I didn't think it was the abbreviation of "Cross"! I think I've seen `xtabs` in other places. But yes if @jamesrkg is happy with the name, we're all set.

jamesrkg:
We're also used the term crosstab. I had actually assume the cr stood for Crunch as well and would have suggested `crosstab` as an alternative but I can see that would be redundant now. `crtabs` is fine for me.

Issue #303, opened 2018-10-16
Project.children


jamesrkg:
We need a way to iterate over the contents of a Project as we can with Dataset, using the new Folders paradigm.

For Datasets we have this: https://github.com/Crunch-io/scrunch/wiki/User-Reference#children

So we'd want something like:

```python
for child in project.children:
    if isinstance(child, Dataset):
        print("I am dataset %s" % child.name)
    elif isinstance(child, Project):
        print("I am a project %s" % child.name)
```

jjdelc:
Work done in https://github.com/Crunch-io/scrunch/pull/304

Issue #298, opened 2018-10-09
Project.delete()


jamesrkg:
Since working with Project.order is now deprecated, we need a way to delete projects. `Project.delete()` seems to be the most consistent way to support this now that we're dealing with nested projects.

jjdelc:
Work done in https://github.com/Crunch-io/scrunch/pull/304

PR #304, opened 2018-10-20
.children and .delete()

jjdelc:
Tackles #298 and #303

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -320,11 +320,11 @@ class Members:
 
     def _validate_member(self, member):
         """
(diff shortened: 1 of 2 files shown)


jamesrkg:
@jjdelc i did some tests with this and noticed that iterating over `children` will return a recently deleted project. I guess `delete` needs to refresh the project resource?

jjdelc:
@jamesrkg You mean that you have `parent` and delete `child` project, then `parent` will still try to return `child`?

I think that then I need `.children` refresh before starting the iteration, to always do it over a fresh list. Thanks for pointing this out.

PR #305, opened 2018-10-22
Improve `_get_dataset(…)` function so it makes less requests to crunch

dmonroy:
Improve `_get_dataset(…)` function so it makes less requests to crunch. If `dataset` parameter is a uuid, try loading it directly first.

When doing a `get_dataset(dsid)` it was first loading the `/datasets/` catalog, then `/datasets/{dsid}/`. It makes no sense to load the datasets catalog, specially when the user performing that task has many datasets as it makes the request really slow.

With this PR it validates if the given dataset reference is a valid uuid, and if so, it loads the dataset directly from dataset endpoint, without loading the catalog.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -21,7 +21,7 @@ from scrunch.folders import DatasetFolders
 from scrunch.helpers import (ReadOnly, _validate_category_rules, abs_url,
                              case_expr, download_file, shoji_entity_wrapper,
                              subvar_alias, validate_categories, SELECTED_ID,
-                             NOT_SELECTED_ID, NO_DATA_ID)
+                             NOT_SELECTED_ID, NO_DATA_ID, validate_uuid)
 from scrunch.order import DatasetVariablesOrder, ProjectDatasetsOrder
 from scrunch.subentity import Deck, Filter, Multitable
 from scrunch.variables import (combinations_from_map, combine_categories_expr,
@@ -132,7 +132,7 @@ def _get_dataset(dataset, connection=None, editor=False, project=None):
                 "Authenticate first with scrunch.connect() or by providing "
                 "config/environment variables")
     root = connection
-    root_datasets = root.datasets
+    shoji_ds = None
     # search on project if specifed
     if project:
         if isinstance(project, six.string_types):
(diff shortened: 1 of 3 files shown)


scrunch/datasets.py
```diff
@@ -141,23 +141,30 @@ def _get_dataset(dataset, connection=None, editor=False, project=None):
...
             shoji_ds = project.get_dataset(dataset).resource
     else:
-        # search by dataset name
-        try:
-            shoji_ds = root_datasets.by('name')[dataset].entity
-        except KeyError:
-            # search by dataset id
+        # Check if dataset reference is a valid uuid, if so, try to load it
+        # directly
+        if validate_uuid(dataset):
```
jjdelc: Do not rely on the ID to be a UUID, it may not be in the future (if we move towards shorter IDs). I suggest you just always try to get by ID first, and if it doesn't work then do the other strategy.

scrunch/datasets.py
```diff
@@ -141,23 +141,30 @@ def _get_dataset(dataset, connection=None, editor=False, project=None):
...
-                        root.catalogs.datasets, '{}/'.format(dataset))
-                    shoji_ds = root.session.get(dataset_url).payload
-                except Exception:
-                    raise KeyError(
-                        "Dataset (name or id: %s) not found in context."
-                        % dataset)
+                dataset_url = urljoin(
+                    root.catalogs.datasets, '{}/'.format(dataset))
+                shoji_ds = root.session.get(dataset_url).payload
+            except Exception:
```
jjdelc: Catch the right exception, will be a `pycrunch.ClientError` with `status` 404, just then `pass` it. Else continue raising it in case something else went wrong. You only want to ignore the exception if it is a 404 Not Found,

scrunch/datasets.py
```diff
@@ -141,23 +141,30 @@ def _get_dataset(dataset, connection=None, editor=False, project=None):
         else:
             shoji_ds = project.get_dataset(dataset).resource
```
jjdelc: If `project.get_dataset()` returns the `Dataset` instance, why you're fetching its `.resource` again and making another `Dataset` instance?

dmonroy: It was like that before my intervention with this PR, all I can guess is that `scrunch.datasets.get_dataset(...)` expects the shoji document, and not the Dataset instance.

I also see that `scrunch.datasets.get_dataset(...)` gonna be deprecated, so perhaps we can make it behave better on the newer methods.

jjdelc: Ok, it's fine to do it in another ticket.

scrunch/helpers.py
```diff
@@ -1,5 +1,6 @@
 import requests
 import six
+import uuid
```
jjdelc: You don't need this import

scrunch/tests/test_utilities.py
```diff
@@ -119,7 +127,7 @@ def test_get_dataset_from_project_no_name(self, session):
         shoji_entity = {
             'element': 'shoji:entity',
             'body': {
-                'id': '123456',
+                'id': 'b2c4c6b7d3a94e58937b23c1fed1b65e',
```
jjdelc: Factor this long string into a variable to make it more evident in the test that you're using the exact same string here and in the other places.

PR #312, opened 2018-10-29
Members edits

jamesrkg:
TODO: Update tests.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -309,19 +309,15 @@ class Members:
         :return: A list of members of the Entity as strings. A member
             can be a User or a Team. Returns ['user1@example.com', 'Team A']
         """
-        if permissions:
-            members = {'edit': [], 'view': []}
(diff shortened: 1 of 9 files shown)


scrunch/datasets.py
```diff
@@ -304,18 +304,25 @@ class Members:
...
+                group = 'edit' if edit else 'view'
+                members[group].append(name)
+        else:
+            members = []
+            for member in self.resource.members.index.values():
+                # members can be users or teams
+                user = member.get('email')
+                if not user:
+                    user = member.get('name')
+                members.append(user)
```
jamesrkg: @mathiasbc are you aware of a situation where a user might not have an email?

mathiasbc: @jamesrkg: a project won't have an email, that is why below we grab it's name

jamesrkg: @mathiasbc of course, yes - thanks.

Issue #63, opened 2017-02-13
(derived) Support "else" case for new categoricals 


jamesrkg:
Extend support for the Crunch API's `else` case to `create_single_response` and `create_multiple_response`.

For example `'id': 6`, below:

```python
agerange = dataset.create_categorical([
    {'id': 1, 'name': 'Underage', case: 'age < 18'},
    {'id': 2, 'name': 'Millenials', case: 'age > 18 and age < 25'},
    {'id': 3, 'name': 'Gen X', case: 'age < 35 and age >= 25'},
    {'id': 4, 'name': 'Grown ups', case: 'age < 60 and age >= 35'},
    {'id': 5, 'name': '60+', case: 'age >= 60', 'numeric_value': 100},
    {'id': 6, 'name': 'Unknown', case: 'else'}
], alias='agerange', name='Age range', multiple=False)
```

jamesrkg:
For `categorical` it looks like the the final ELSE is supported here:

http://docs.crunch.io/feature-guide/feature-deriving.html?highlight=else#case-statements

I don't think an example of the ELSE is actually given in that snippet, though. If I read it correctly, _"You may include an extra value for the case when none of the boolean expressions match"_ means `derivation.args` would have a fourth item `{"value": 4}` in order to code non-conforming rows to `Other`?

Given what we've done for #286 we'll also need to consider how this could support derived `multiple_response` as well. `derive_multiple_response` should stay in its current raw form so this would only be an extension to `create_categorical(..., multiple=True)`. Since we're talking about a series of case statements for distinct subvariables I'm fairly sure scrunch will need to stitch all the `cases` together with `not X and not Y and not ...` etc., so that the following:

```python
var = ds.create_categorical(
    name='Preferred soda',
    alias='soda',
    description='Which drinks do you prefer?',
    multiple=True,
    categories=[
        {'id': 1, 'name': 'Coke', 'case': 'q1 in [1]', 'missing': 'missing(q1)'},
        {'id': 2, 'name': 'Pepsi', 'case': 'q1 in [2]', 'missing': 'missing(q1)'},
        {'id': 4, 'name': 'Coke or Pepsi', 'case': 'q1 in [1,2]', 'missing': 'missing(q1)'},
        {'id': 3, 'name': 'Other brands', 'case': 'q1 in [3, 4, 5, 6, 95]', 'missing': 'missing(q1)'},
[...]

PR #321, opened 2018-11-12
Feature/else case 63

mathiasbc:
This is currently a WIP, currently working for `create_categorical(multiple=False)` only. Need to clarify some things for Multiple Response types


mathiasbc:
closes #63

Issue #325, opened 2018-11-15
create_categorical missing_case not applied when cases overlap


jamesrkg:
In the following example the `missing_case` is not mutually exclusive to the given `case` statements:

```
var = ds.create_categorical(
    alias='test_3',
    name='test_3',
    description='test_3',
    multiple=True,
    missing_case='gender == 1',
    categories=[
        {'id': 1, 'name': '1+2', 'case': 'household_income_recoded in [1, 2]'},
        {'id': 2, 'name': '1+2+3', 'case': 'household_income_recoded in [1, 2, 3]'},
    ]
)
```

This results in the following `subvariables` for `derived_multiple_response`:

```
[
  {
    "cases": {
      "1": "household_income_recoded in [1, 2]",
      "2": "not (household_income_recoded in [1, 2]) and not (gender == 1)",
      "-1": "gender == 1"
    },
    "id": 1,
    "name": "1+2"
  },
  {
    "cases": {
      "1": "household_income_recoded in [1, 2, 3]",
      "2": "not (household_income_recoded in [1, 2, 3]) and not (gender == 1)",
      "-1": "gender == 1"
    },
    "id": 2,
    "name": "1+2+3"
  }
]
```

However the result of this is unexpected The `selected` category (`1`) is not being limited by `and not (gender == 1)`:

```
              else_3           
                 1+2 1+2+3    N
gender Kvinde    135   195  135
       Mand      132   206  247
       N         267   401  382
```

In the above we would expect the `Kvinde` row to be all `0`s.

This can be corrected by adding `and not (missing_case)` by default to all `selected` AND `not selected` cases:

```
[
  {
    "cases": {
      "1": "(household_income_recoded in [1, 2]) and not (gender == 1)",
      "2": "not (household_income_recoded in [1, 2]) and not (gender == 1)",
      "-1": "gender == 1"
    },
    "id": 1,
    "name": "1+2"
  },
  {
    "cases": {
      "1": "(household_income_recoded in [1, 2, 3]) and not (gender == 1)]",
      "2": "not (household_income_recoded in [1, 2, 3]) and not (gender == 1)",
      "-1": "gender == 1"
    },
    "id": 2,
    "name": "1+2+3"
  }
]
```

Which results in the expected:

```
              gender test_3           
               Total    1+2 1+2+3    N
gender Kvinde    260      0     0    0
       Mand      247    132   206  247
       N         507    132   206  247
```

[...]

PR #326, opened 2018-11-15
Dealing with missing_case in else statement for SELECTED category, cl…

mathiasbc:
…oses #325

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1574,7 +1574,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
 
                 if 'missing_case' in sv:
                     selected_case = '({}) and not ({})'.format(sv['case'], sv['missing_case'])
-                    not_selected_case = 'not ({})'.format(selected_case)
+                    not_selected_case = 'not {}'.format(selected_case)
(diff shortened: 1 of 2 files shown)


scrunch/datasets.py
```diff
@@ -1573,12 +1573,13 @@ def create_categorical(self, categories, alias, name, multiple, description='',
                 sv['case'] = get_else_case(sv['case'], categories)
 
                 if 'missing_case' in sv:
-                    not_selected_case = 'not ({}) and not ({})'.format(sv['case'], sv['missing_case'])
+                    selected_case = '({}) and not ({})'.format(sv['case'], sv['missing_case'])
+                    not_selected_case = 'not ({})'.format(selected_case)
```
jamesrkg: Small error in my suggested fix. We need to remove the parentheses around the insertion of `selected_case`.

Instead of this:

`not_selected_case = 'not ({})'.format(selected_case)`

It should be this:

`not_selected_case = 'not {}'.format(selected_case)`

mathiasbc: oh I see. This was my mistake I added the parentheses thinking we wanted to wrap it all. removing and updating tests now.

PR #336, opened 2018-12-12
derive_weight now allows to include more than 1 variable to raking

mathiasbc:


xbito:
I guess tests are coming though right? So no Merge yet? @mathiasbc

mathiasbc:
@xbito: yes, I need to write tests I was just waiting on your approval that we still want it this way.

PR #338, opened 2019-01-29
adding replace_from_csv method to streaming_dataset, since we are bas…

### LINKED ISSUE #288: ds.replace_from_csv
The idea here is as-per `replace_values` except that a CSV is consumed containing a column of ids paired to the dataset and 1+ additional columns headed by sub/varaible aliases. The method would then match on the ids to overwrite those columns with this indicated data.

For example, given the CSV:

```
id	gen	age	reg
43	1	38	8
31	1	22	2
41	2	42	5
82	1	25	5
26	2	19	3
92	2	29	4
62	1	51	9
```

Something like:

```
ds.replace_from_csv(match_on='id', data=path_csv, sep=',')
```

Resulting in all of the data from the CSV overwriting whatever data was in those variables, for those ids.

@jjdelc can you suggest the best approach for tackling this one using the fewest possible hits against the API. Could it be reduced to 1 call per id? 1 per variable? Can the API take all of this in one go? If 1 call per id per variable is required then internally this could rely on `var.replace_values`, each time filtering on the id.

I've framed this as `replace_from_csv` but it could also be `replace_from_dataframe` (didn't realize initially that pandas is already a dependency).

mathiasbc:
…ing the replace on the stream endpoint, closes #288


jjdelc:
This looks incomplete, ideally you'd switch the dataset to streaming, stream the new csv rows and then turn back to the original state right? That should happen inside the `replace_from_csv` function seamlessly for the user.

This is also missing tests.

scrunch/tests/test_datasets.py
```diff
@@ -250,6 +256,17 @@ def test_edit_dataset(self):
...
         return expr
 
+    def test_replace_from_csv(self):
+        ds_mock = self._dataset_mock()
+        ds = StreamingDataset(ds_mock)
+        assert ds.resource.body.get('streaming') == 'streaming'
+        file = StringIO()
+        file.write("id, age\n1, 15")
+        file.seek(0)
+        with pytest.raises(AttributeError):
```
jjdelc: What `AttributeError` is being raised here? Why?

scrunch/tests/test_datasets.py
```diff
@@ -250,6 +256,17 @@ def test_edit_dataset(self):
...
+            assert ds.replace_from_csv(file, push_rows=False) == 1
+        assert ds.resource.body.get('streaming') == 'streaming'
```
jjdelc: Also assert that the rows have been streamed in, or mock that the `push_rows` function was called with the rows you wanted`. Right now this test isn't protecting the implementation.

scrunch/datasets.py
```diff
@@ -2365,6 +2369,37 @@ def replace_values(self, variables, filter=None, literal_subvar=False):
...
+        )
+        for chunk in df_chunks:
+            # This is a trick to get rid of np.int64, which is not
+            # json serializable
+            stream = chunk.to_json(orient='records')
+            stream = json.loads(stream)
+            importer.stream_rows(self.resource, stream)
+            if push_rows:
+                streaming_ds.push_rows(chunksize)
+        self.edit(streaming=streaming_state)
```
jjdelc: Here you would be switching the streaming state back to "`no`". I am not sure if it is possible to make this change. Can you confirm? I recall there being some limitations about which way the streaming states of datasets could go.

jjdelc: I am checking, and you can only go back to "`no`" IFF there are 0 pending rows. So if you didn't push rows, you cannot make your ds go back to "no" and will be stuck as streaming until you push them. When do you need to not push them? Why is the flag there?

scrunch/tests/test_datasets.py
```diff
@@ -250,6 +256,17 @@ def test_edit_dataset(self):
     def process_expr_side_effect(self, expr, ds):
         return expr
 
+    def test_replace_from_csv(self):
+        ds_mock = self._dataset_mock()
+        ds = StreamingDataset(ds_mock)
+        assert ds.resource.body.get('streaming') == 'streaming'
```
jjdelc: Also test that a dataset that is with `streaming == no` gets back to its original state.

scrunch/datasets.py
mathiasbc: Yes I realized we cannot move the streaming flag when there is pending batches, so I will force to push rows. Ultimately if this change fails, the user will be get an Exception with the message.

jamesrkg:
I get the following (trying on a mutable dataset):

```
----> 1 ds.replace_from_csv(path_csv)

C:\ProgramData\Anaconda2\lib\site-packages\scrunch\scrunch\datasets.py in replace_from_csv(self, filename, chunksize)
   2403             # and to allow changing to streaming status back to it's previous
   2404             # state
-> 2405             ds.push_rows(chunksize)
   2406         if streaming_state != 'streaming':
   2407             ds.edit(streaming=streaming_state)

C:\ProgramData\Anaconda2\lib\site-packages\scrunch\scrunch\datasets.py in __getattr__(self, item)
    954             return self.resource.body[item]  # Has to exist
    955         # Default behaviour
--> 956         return object.__getattribute__(self, item)
    957 
    958     def __repr__(self):

AttributeError: 'MutableDataset' object has no attribute 'push_rows'
```

mathiasbc:
@jamesrkg: this is strange, there is a check for that:
```
if streaming_state != 'streaming':
            ds = self.make_streaming()
```

and I tested it with the `get_mutable_dataset` method and it works fine; also the test I added tests that particular case as well: https://github.com/Crunch-io/scrunch/pull/338/files#diff-ae872301ff713c915a36ec3866811bb9R265

Please make sure you have the latest changes and try again, or paste the code, perhaps I'm missing something.

jamesrkg:
@mathiasbc i was up to date, but i now can't reproduce that error.. 

However, I'm working on `f579ac179faf49b0a5ff84fd5c4dd9a6` which happens to have duplicates in the pk column 'identity'. In this case, should the method refuse the request?

jamesrkg:
@mathiasbc also i don't believe this is properly resetting an initially mutable dataset back to that state.

mathiasbc:
I can't think or reproduce a situation where a not streaming Dataset is not reseted back to that state:https://github.com/Crunch-io/scrunch/pull/338/files#diff-ae872301ff713c915a36ec3866811bb9R273

Unless you happen to have an error mid run and the line that sets the streaming state back to normal was never reached.

mathiasbc:
@jamesrkg: I can't see a pk set for that DS: https://app.crunch.io/api/datasets/f579ac179faf49b0a5ff84fd5c4dd9a6/pk/

xbito:
I was looking around to see if we can implement a replace_from_dataframe, and then have a replace_from_csv that just calls it after loading the CSV, but I noticed that the current implementation reads the CSV in chunks if needed. So, I'll leave it as it is, seems like a nice optimization to have.

PR #343, opened 2019-02-04
use `str` in py3 instead of basestring

### LINKED ISSUE #342: basestring is not defined (py3)
the folders modules uses `basestring` which does not exist in python3

skriems:
closes #342


skriems:
Absolutely! Hope that looks better now.

PR #354, opened 2019-03-12
Warn on usage of Dataset.change_owner, favor `Dataset.project` and `D…

jjdelc:
We are going to move away from using the `Dataset.owner` as the way to detect and change the project where the dataset is stored.

The API should favor the usage of `Dataset.project` to get the containing project and `Dataset.move(project)` to change it. Deprecating `Dataset.owner` and `Dataset.change_owner`.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1046,8 +1046,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
             if not isinstance(project, Project):
                 project = get_project(project)
 
-        project.move_here([self])
-        self.resource.refresh()
+        project.move_here([self])  # This performs .resource.refresh()
 
     @property
     def settings(self):
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -6,7 +6,7 @@ import logging
 import os
 import re
 import sys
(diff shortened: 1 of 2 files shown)


scrunch/datasets.py
```diff
@@ -1017,24 +1021,32 @@ def change_owner(self, user=None, project=None):
...
                 project = get_project(project)
-            owner_url = project.url
-
-        if not owner_url:
+            self.move(project)
+        else:
             raise AttributeError("Can't set owner")
 
-        self.resource.patch({'owner': owner_url})
+    def move(self, project):
```
jjdelc: I'm not sure if we want to have `Dataset.move`, when we already have `Project.move_here`, The location of a dataset is not its responsibility, this could make the code simpler and less entangled with `Project` code.

jjdelc: My idea was to remove the project moving code (regardless of the function being a setter or a method) out of the Dataset class. Since the placing of datasets is a concern of projects and not of datasets, but between both options I much prefer the `move()` function as it is very explicit in the code that it is _moving_ the dataset to a particular location. A setter in this case wouldn't communicate the same semantics.

PR #374, opened 2019-09-12
when creating a variable with create_variable() and passing subvariab…

mathiasbc:
…les with id and alias, respect those in favor of default values generated by Crunch

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -1708,8 +1708,6 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                 subrefs = {'name': item['name']}
                 if item.get('alias'):
                     subrefs['alias'] = item['alias']
-                if item.get('id'):
-                    subrefs['id'] = item['id']
                 payload['subreferences'].append(subrefs)
         if values:
             payload['values'] = values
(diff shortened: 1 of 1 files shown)


alextanski:
Great. Let's remove code for getting the `id` for clearity. Crunch derives it from the url of the subreference autonomously.

scrunch/datasets.py
```diff
@@ -1703,9 +1703,14 @@ def create_variable(self, var_type, name, alias=None, description='',
...
         if subvariables:
-            payload['subreferences'] = [
-                {'name': item['name']} for item in subvariables
-            ]
+            payload['subreferences'] = []
+            for item in subvariables:
+                subrefs = {'name': item['name']}
+                if item.get('alias'):
+                    subrefs['alias'] = item['alias']
+                if item.get('id'):
```
alextanski: Let's remove that bit after all. Sorry!

alextanski: I was just making a point about the clarity of the code, i.e. prevent other users from expecting the `id` is actually used by Crunch when included in the payload because Crunch will ignore it regardless.

mathiasbc: I will proceed to remove it then, it doesn't make sense to have it there if Crunch will ignore it.

PR #376, opened 2019-10-31
StreamingDataset and MutableDataset have now deprecated warning messa…

mathiasbc:
…ges. They still exists but their behaviour is directly inherited from the Dataset class, which makes them identical at this point


jamesrkg:
How does the scrunch.dataset.LOG work? We also use a logger but the deprecation warning is not coming up. It might be better to raise a DeprecationWarning in a simpler way to ensure its visible to all users.

jamesrkg:
I think `Dataset` would benefit from:

```python
@property
def streaming(self):
    return self.resource.body.get('streaming', 'no')
```

PR #381, opened 2020-05-11
for recodes test, column order doesn't matter.

percious:


integration/test_recodes.py
```diff
@@ -111,9 +111,30 @@ def test_recodes(self):
...
         dataset.export(output.name)
-        result = [l.strip() for l in output.read().strip().split('\n')]
-        expected = RECODES_CSV_OUTPUT.split('\n')
-        # Rows are unordered under streaming conditions
-        self.assertEqual(sorted(result), sorted(expected))
+
+        # put the data into columns
+        actual = {}
+        line = output.readline()
+        headers = line.strip().split(',')
```
jjdelc, line 118: Better to use python's CSV reader on the exported file and on our fixture and compare, saves us lots of shaky parsing code.

percious: i don't understand why this is a requirement when the previous rendition also did not use a csv parser.

PR #385, opened 2020-07-13
BaseDataset.replace_values should not include parameters when executing a POST 

AlasdairE:
When a `filter` is passed in `BaseDataset.replace_values()` ...

- `process_expr` is called
- `process_expr` calls `get_dataset_variables`
- `get_dataset_variables` calls table endpoint, setting `limit=0` parameter on the table url
- back to `replace_values`, it then executes post on table resource, which now is set to have `?limit=0`

This pull request will clear the remove the parameters from the table resource url, if filter is passed, before executing a POST, thus avoiding the 404 errors

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2386,9 +2386,10 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                     payload['variables'][self[alias].id] = {'value': val}
         if filter:
             payload['filter'] = process_expr(parse_expr(filter), self.resource)
-            # Remove query parameters from table url
-            table = self.resource.table
-            table.self = table.self[:table.self.find('?')]
+
+        # Remove query parameters from table url
+        table = self.resource.table
+        table.self = table.self[:table.self.find('?')]
 
         resp = self.resource.table.post(json.dumps(payload))
         if resp.status_code == 204:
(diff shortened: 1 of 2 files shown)


scrunch/datasets.py
```diff
@@ -2386,6 +2386,9 @@ def replace_values(self, variables, filter=None, literal_subvar=False):
                     payload['variables'][self[alias].id] = {'value': val}
         if filter:
             payload['filter'] = process_expr(parse_expr(filter), self.resource)
+            # Remove query parameters from table url
+            table = self.resource.table
+            table.self = table.self[:table.self.find('?')]
```
dmonroy, line 2391: Do we need it only `if filter`?... I know the `limit` query parameter is set when `process_expr` (or another function it depends on) is called, but there's also the possibility that somewhere else, before `replace_values` is called, has also loaded the table, hence the resource's self member will include the `limit` query parameter.

I may be wrong, just consider the possibility and move outside of the `if filter` if you consider it is needed @AlasdairE .

AlasdairE: Yup, I have moved that outside the `if` now:
https://github.com/Crunch-io/scrunch/pull/385/files#diff-10a14081413b0535e4d0097c2ad71a58R2390

dmonroy: LGTM!

PR #386, opened 2020-07-21
Add `fill` support

jjdelc:
Add support for categorical `fill` variables via the `create_fill_values` method


jjdelc:
Added docs https://github.com/Crunch-io/scrunch/wiki/User-Reference#create_fill_values

PR #388, opened 2020-10-13
Add support for editing the `date` field on categoricals

### LINKED ISSUE #387: Let me set date field on Categories
A `date` field was added to categorical variables to better support time series with surveys. It may be as simple as adding it to the list of `mutable_attributes` with a test or two. 
It accepts ISO8601 (`2020-10-07`) including ‘reduced accuracy’ forms such as `2020`, `2020-10`, and `2020-W41`, with server-side validation that dates on categories are unique and that all valid categories have a date set. (Unlike numeric values, which may be null). Categories should be arranged in chronological order. (If you are adding historical data, you may need to adjust category order first.)

alejandro-rivera:
Fixes #387.

### DIFF
--- a/scrunch/categories.py
+++ b/scrunch/categories.py
@@ -6,7 +6,7 @@ from scrunch.helpers import ReadOnly
 
 
 class Category(ReadOnly):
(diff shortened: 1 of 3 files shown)


jjdelc (approved):
looks good

PR #393, opened 2021-02-04
Backfill w excl

jjdelc:
This PR temporarily removes the exclusion filter on the dataset during the backfill process.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -3486,8 +3486,6 @@ class BackfillFromCSV:
             "filter": self.rows_expr
         }
         with NoExclusion(self.dataset) as ds:
-            print("Current excl")
-            print(ds.resource.exclusion)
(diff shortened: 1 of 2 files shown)


scrunch/datasets.py
```diff
@@ -3460,7 +3485,10 @@ def backfill(self):
             "variables": variables_expr,
             "filter": self.rows_expr
         }
-        self.dataset.resource.table.post(update_expr)
+        with NoExclusion(self.dataset) as ds:
+            print("Current excl")
+            print(ds.resource.exclusion)
```
xbito, line 3490: Should we remove those couple of prints?

xbito (changes requested):

scrunch/datasets.py
jjdelc: Done

PR #394, opened 2021-02-09
Make `Dataset.replace_values()` wait for the progress to finish for A…

jjdelc:
…synchronous responses

### DIFF
--- a/integration/test_dataset.py
+++ b/integration/test_dataset.py
@@ -1,0 +1,53 @@
+# coding: utf-8
+
+import os
+from unittest import TestCase
+
+from scrunch import connect
+from scrunch.mutable_dataset import get_mutable_dataset
+
+HOST = os.environ['SCRUNCH_HOST']
+username = os.environ['SCRUNCH_USER']
+password = os.environ['SCRUNCH_PASS']
+
+
+site = connect(username, password, HOST)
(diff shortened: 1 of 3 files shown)


integration/test_dataset.py
```diff
@@ -0,0 +1,53 @@
...
+                    "type": "numeric",
+                    "values": [1, 2, 3, None, 5],
+                }
+            )
+        ).refresh()
+        scrunch_dataset = get_mutable_dataset(ds.body.id, site)
+        resp = scrunch_dataset.replace_values({
+            "my_var": 4
+        }, filter="missing(my_var)")
+        if resp is not None:
```
jjdelc, line 40: As an HTTP client I can't force-request an asynchronous response. This test evaluates synchronously bc of the size of the input.

scrunch/datasets.py
```diff
@@ -2537,6 +2537,7 @@ def replace_values(self, variables, filter=None, literal_subvar=False):
         if resp.status_code == 204:
             LOG.info('Dataset Updated')
             return
```
jjdelc, line 2539: Here I haven't changed the return value for `replace_values()` although I think we should always return the response instead of sometimes `None` and sometimes the response.
Making this change may break existing code that's relying on the response being None in some cases?

PR #396, opened 2021-03-09
Dataset scripts

jjdelc:
Adds support for executing Crunch Automation scripts on Scrunch under:
```
dataset.scripts.execute()
```


xbito (approved):

xbito:
@jjdelc we still use python 3.4 in some places. But, we might just want to deprecate those tests anyway.

PR #406, opened 2021-05-07
Fix table url that contains query string in BaseDataset.replace_values

AlasdairE:
This pull request https://github.com/Crunch-io/scrunch/pull/385 aimed to take care of the error

```
<Response [404]>, 'https://app.crunch.io/api/datasets/<dataset id>/table/?limit=0', {'status': '404 Not Found', 'message': 'Nothing matches the given URI'}
```

when using the `BaseDataset.replace_values` method. This seems not to work since the post is not done on the updated url.

Since the `limit=0` is the default for the table endpoint in crunch, we should remove it in `get_dataset_variables` and then we won't have a query string on the url in `BaseDataset.replace_values` when doing the post.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2535,10 +2535,6 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         if filter:
             payload['filter'] = process_expr(parse_expr(filter), self.resource)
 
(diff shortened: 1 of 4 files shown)


jjdelc:
I'll be closing this PR, with the last updates on Crunch API this fix will not be necessary.

PR #412, opened 2021-06-24
Don't send null date if unnecessary editing categories

jjdelc:


xbito:
@jjdelc the tests are failing, though I don't think it is related, all ok to merge anyway?

jjdelc:
Addressed the other tests failures, indeed they were introduced in a different PR.

PR #420, opened 2021-12-07
improve join speed

bhagya9631:


scrunch/mutable_dataset.py
```diff
@@ -100,8 +100,9 @@ def join(self, left_var, right_ds, right_var, columns=None,
                 ]
             }
             # add the individual variable columns to the payload
-            for var in columns:
-                var_url = right_ds[var].url
+            alias_list = right_ds.resource.variables.by("alias")
+            var_urls = ['{}variables/{}/'.format(right_ds.resource.self, alias_list[alias]['id']) for alias in columns]
```
jjdelc, line 104:
```suggestion
            var_urls = [alias_list[alias].entity_url for alias in columns]
```

In pycrunch, we should always think it's a smell if we're interpolating strings to construct URLs, there's usually another way to get them directly from the API.

scrunch/mutable_dataset.py
```diff
@@ -100,8 +100,9 @@ def join(self, left_var, right_ds, right_var, columns=None,
...
+            for var_url in var_urls:
```
jjdelc, line 105:
```suggestion
            var_urls = {var_url: {"variable": var_url} for var_url in var_urls}
            payload['body']['args'][0]['map'] = var_urls
```

jjdelc (changes requested):
We shouldn't extrapolate strings to build the URL

jjdelc (approved):

PR #421, opened 2021-12-09
Ds join better control for wait

bhagya9631:


scrunch/mutable_dataset.py
```diff
@@ -113,7 +115,8 @@ def join(self, left_var, right_ds, right_var, columns=None,
         progress = self.resource.variables.post(payload)
         # poll for progress to finish or return the url to progress
         if wait:
```
jjdelc, line 117: we should change this to `wait or timeout` in case they stop using `wait` and start using `timeout` as the preferred argument. Having two arguments that do something similar will become confusing.

jjdelc (changes requested):

jjdelc (approved):

PR #429, opened 2022-02-12
Do not read email from session

jjdelc:


jjdelc:
@xbito https://github.com/Crunch-io/scrunch/pull/429/commits/e24d8e7f4918a1f8ae2cdcc3321f04c1b15124a5 this commit makes all the calls to `change_editor` do one less request.

PR #433, opened 2022-07-14
Add `.views` API to handle Crunch views

jjdelc:


scrunch/tests/test_datasets.py
```diff
@@ -1997,7 +1997,6 @@ def prepare_ds(self):
         dataset_url = 'http://host/api/projects/abc/'
         variables_url = "http://host/api/projects/abc/variables/"
         new_var_url = "http://host/api/projects/abc/variables/123/"
-        hier_url = "http://host/api/projects/abc/variables/hier/"
```
jjdelc, line 2000: Now accessing this endpoint is lazy, so it is not a request done during initialization. It was here as part of the needed fixtures to init a dataset

scrunch/datasets.py
```diff
@@ -952,11 +953,22 @@ def _reload_variables(self):
...
         self._set_catalog()
         self._vars = self._catalog.index.items()
-        order = self._catalog.hier
+        self._order = None
 
-        # The `order` property, which provides a high-level API for
-        # manipulating the "Hierarchical Order" structure of a Dataset.
-        self.order = DatasetVariablesOrder(self._catalog, order)
+    @property
+    def order(self):
```
bhagya9631, line 959: is there a testcase for this?

scrunch/views.py
```diff
@@ -0,0 +1,49 @@
+# coding: utf-8
+
+from scrunch.helpers import shoji_entity_wrapper
+
+
+class DatasetViews:
+    def __init__(self, dataset_resource):
+        self.dataset_resource = dataset_resource
+
+    def create(self, name, columns):
```
rchacon1, line 10: do u plan on adding `rows` at later time?

rchacon1: and destination `project`?

rchacon1:
@jjdelc can we get the python 2.7 build to pass?

scrunch/views.py
jjdelc: tried to keep it at a minimum this time, just to make sure we had support. I'm sure they'll request it and we can have smaller PRs for those.

PR #437, opened 2022-12-01
Crunch accounts

jjdelc:


scrunch/connections.py
```diff
@@ -0,0 +1,106 @@
...
+    if connection_kwargs:
+        return connect(**connection_kwargs)
+
+    # try reading from .ini file
+    config = configparser.ConfigParser()
+    config.read(file_path)
+    try:
+        site = config.get('DEFAULT', 'CRUNCH_URL')
+        connection_kwargs["site_url"] = site
+    except Exception:
```
rchacon1, line 72: Q: Is there something more specific you are concerned about? `KeyError` perhaps.

scrunch/tests/test_accounts.py
```diff
@@ -0,0 +1,108 @@
...
+    def test_run_script(self):
+        session = self.make_session()
+
+        response = Response()
+        response.status_code = 204
+
+        session.add_post_response(response)
+        current_act = Account.current_account(session.root)
+
+        current_act.run_script("NOOP;")
```
rchacon1, line 67: `AttributeError: 'Account' object has no attribute 'run_script'`

rchacon1:
Q: Can we get the automated build to pass? They are either failed or cancelled

scrunch/connections.py
jjdelc: you're right. I simply moved this to another file and didn't check the code. Fixing this

integration/test_accounts.py
```diff
@@ -0,0 +1,46 @@
...
+    "element": "shoji:entity",
+    "body": b
+}
+
+
+HOST = os.environ["SCRUNCH_HOST"]
+username = os.environ["SCRUNCH_USER"]
+password = os.environ["SCRUNCH_PASS"]
+
+site = connect(username, password, HOST)
```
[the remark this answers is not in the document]
jjdelc: I change this to use my API key on local test, but the integration tests are using username and password on jenkins. That's how all the other integration tests are working currently.

scrunch/accounts.py
```diff
@@ -0,0 +1,72 @@
+# coding: utf-8
+
+import pycrunch
+from scrunch.helpers import shoji_view_wrapper
+from scrunch.scripts import ScriptExecutionError
+from scrunch.connections import _default_connection
+from scrunch.datasets import Project
+
+
+class AccountProjects:
```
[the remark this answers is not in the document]
jjdelc: We decided against that. This reads from the `/api/account/projects/` API that we recently modified from returning all projects to only toplevel folders. This is an API decision. This allows an account admin to explore and navigate their tree. Returning everything flat can result in a too big catalog and not what they need.

Also, this endpoint works as the base to run system scripts at "account" level, so everything here is used as the top folders you can run things against. Listing everything would confuse about which are the top level folders and which subfolders need to be addressed by a nested path.

scrunch/tests/test_accounts.py
jjdelc: Eek, left over from rename. It's now called `.execute` changed

rchacon1 (approved):
:shipit:

PR #443, opened 2023-03-03
Fix reading floats as category IDs

jjdelc:


aless10:
@jjdelc I think that this has been fixed in #459. can we close it?

PR #444, opened 2023-06-21
fix task error

gergness:
Can't tell if there's some reason why this hasn't been fixed yet, but after returning to python dev after a month it seemed crazy we were still getting these flakes.

See https://master.ci.crint.net/blue/organizations/jenkins/Crunch-io%2Fzoom/detail/PR-7207/2/pipeline/ for example


jjdelc (approved):
You're godsent! I just had a ticket to do this! lol

PR #452, opened 2024-04-25
use API key instead of password

rchacon1:


aless10:
@jjdelc @rchacon1 since I recently refactored the integration tests, can I pick up this branch and work on this feature?

rchacon1:
> @jjdelc @rchacon1 since I recently refactored the integration tests, can I pick up this branch and work on this feature?

fine with me. i lost track of this. i don't recall what the workaround was.

jjdelc:
YEs, that would be great. Thanks!

rchacon1:
there is a second user that is used in `integration/test_projects.py` which authn with user/pass. We first need to create a second api_key to be used by this second user.

PR #454, opened 2024-06-06
Strict long form

jjdelc:


scrunch/tests/test_expressions.py
```diff
@@ -1590,6 +1623,125 @@ def test_transform_alias_to_var_id(self):
...
+                },
+                {
+                    'function': 'in',
+                    'args': [
+                        {'variable': "%ssubvariables/003/" % var_url},
+                        {'value': [1]}
+                    ],
+                }],
+            }],
+        }
```
aless10, line 1684: I'm still trying to understand what any and all do. But this test feels wrong to me. So here we have 
`(var0001 has value in [1]) or (var0002 has value in [1] or var0003 has value in [1])`
I would expect this to be
``(var0001 has value in [1]) or (var0002 has value in [1]) or (var0003 has value in [1])``
What do you think?

PR #445, opened 2023-08-10
Update test-and-deploy.yaml

arlind87:

PR #455, opened 2024-06-11
add tests on filter with datetime variables

aless10:
Following slack conversation: https://crunch-io.slack.com/archives/C072AKYD58X/p1717521227887949

### DIFF
--- a/integration/test_dataset.py
+++ b/integration/test_dataset.py
@@ -310,20 +310,10 @@ class TestDatasetMethods(TestCase):
         # affected by the exclusion filter, the result should be only one record.
         filter_value = "2024-03-15T00:00:00.393"
         try:
-            resp = scrunch_dataset.append_dataset(
+            scrunch_dataset.append_dataset(
                 scrunch_dataset_to_append,
                 filter="my_datetime_var > '{}'".format(filter_value)
             )
-            assert resp['body']['filter'] == {
-                'args': [
-                    {
-                        'variable': datetime_append_var['self']
-                    }, {
-                        'value': filter_value
-                    }
-                ],
-                'function': '>'
-            }
             data = ds.follow("table", "limit=20")['data']
             datetime_values = data[datetime_var.body.id]
             # We should have 5 (original dataset) + 1 (filtered in append_dataset)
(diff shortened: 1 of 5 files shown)


integration/test_dataset.py
```diff
@@ -56,6 +56,62 @@ def test_replace_values(self):
...
+                    {
+                        'variable': datetime_var['self']
+                    }, {
+                        'value': filter_value
+                    }
+                ],
+                'function': '>'
+            }
+        else:
+            pass
```
jjdelc, line 110: What does it mean if `resp` is None?

integration/test_dataset.py
```diff
@@ -56,6 +56,62 @@ def test_replace_values(self):
...
+
+        ds.delete()
+        ds_to_append.delete()
```
jjdelc, line 113: Can you assert that the new rows are available after the append?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py27,py36,py37,py38,py39,py311{py27,py36,py37,py38,py39,py311}-pandas
+envlist = py27,py36,py37,py38,py39,{py27,py36,py37,py38,py39,py311}-pandas
```
jjdelc, line 2: Why was this removed from the list?

aless10: I just renamed it, because right now it creates an environment called `py311py27-pandas` (`py311py27-pandas create: /home/alessio/crunch/scrunch/.tox/py311py27-pandas`)

integration/test_dataset.py
aless10: mmh, I just reused the test pattern I found in another one. I can just remove it. Thanks

aless10:
@jjdelc I'm also having issue installing the project locally, because the pycrunch dependency is 0.5.5 from the setup.py file while in the requirements it will get the latest one. Do we need to update the pycrunch version also from the setup.py file?

integration/test_dataset.py
aless10: I think I added the test, but the behavior is not so easy to understand. In fact, when I fetch the dataset after the append call, there are multiple null values, and the only ones that are not null are the one that are correctly filtered. 
Looking at the test and the result in superadmin, I have this in the variable summary:

```
{
  "count": 7,
  "min": "2024-04-03T20:00:52.999",
  "max": "2024-06-03T20:00:52.123",
  "valid_count": 2,
  "missing_count": 5,
  "missing_frequencies": [
    {
      "value": {
        "?": -1
      },
      "count": 5
    }
  ],
```

So, in the test I'm creating a new variable with 5 values. then, I append it and in the resulting dataset I have 7 values: 5 are non valid and 2 valid. I'm not expecting this from this api, but maybe this is the expected behavior

aless10:
Also, I made some changes on the github action file to have a green CI for py2.7, I see now that there is already a PR for this https://github.com/Crunch-io/scrunch/pull/445/files. Let me know if you want me to rebase from that one

aless10:
@jjdelc I run the integration tests against alpha which has the code from the fix https://github.com/Crunch-io/zoom/pull/8913 
The tests pass. So that is a good news

PR #457, opened 2024-06-20
add integration test for `any` in an append source filter

aless10:


integration/test_dataset.py
```diff
@@ -53,8 +56,87 @@ def test_replace_values(self):
...
+            [4, 10, 2, 1, 2],
+            [5, 11, 2, 1, 1],
+            [6, 12, 1, 2, 2]
+        ]
+        Importer().append_rows(ds, ds_rows)
+        Importer().append_rows(ds_to_append, ds_to_append_rows)
+        scrunch_dataset = get_mutable_dataset(ds.body.id, site)
+        scrunch_dataset_to_append = get_mutable_dataset(ds_to_append.body.id, site)
+        try:
+            scrunch_dataset.append_dataset(scrunch_dataset_to_append, filter="newssource.any([1])")
```
jjdelc: If `nressource` is a MR variable, then the argument inside `any` should be one of its subvariables I'm afraid. This should look like `newsource.any([newssource_1])`

integration/test_dataset.py
```diff
@@ -53,8 +56,87 @@ def test_replace_values(self):
...
 
         r = ds.follow("table", "limit=10")["data"]
-        assert r[variable.body["id"]] == [1, 2, 3, 4, 5]
-        ds.delete()
+        try:
+            assert r[variable.body["id"]] == [1, 2, 3, 4, 5]
+        finally:
+            ds.delete()
+
+    def test_append_dataset_any_filter(self):
```
jjdelc:
```suggestion
    def test_append_dataset_any_filter_multiple_response(self):
```
We'll want another test that does somthing similar by testing `.any()` on a regular categorical.

jjdelc (changes requested):
Great tests, they're very clean and easy to understand.

aless10:
@jjdelc I moved the tests on a dedicated module `/integration/test_expressions.py` so that they can be find easier

integration/fixtures.py
```diff
@@ -127,3 +127,10 @@ def mr_in(ds, mr_alias, subvars):
             'column': [subvariables[subvar_alias(mr_alias, sv)].id for sv in subvars]
         }]
     }
+
+
+def as_entity(body):
```
aless10, line 132: I moved this one here because I was having a warning for the lambda that should not be assigned. That said, We have the same function here in pycrunch https://github.com/Crunch-io/pycrunch/blob/9bb073368293f1e84843a53bdc0cfccd1c90787e/src/pycrunch/shoji.py#L25. Should we use that one here?

jjdelc (commented):
Only very minor comments. excellent work

integration/test_expressions.py
```diff
@@ -0,0 +1,370 @@
...
+                [1, 3, -1],
+                [2, 1, 1],
+                [2, 3, 2]
+            ]
+        }))
+        scrunch_dataset = get_mutable_dataset(ds.body.id, self.site)
+        _filter = "categorical_var.any([1])"
+        try:
+            resp = scrunch_dataset.add_filter(name='filter_1', expr=_filter)
+            assert resp.resource.body.entity.body.definition == {
```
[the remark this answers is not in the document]
aless10: Done, thanks for the tip

aless10:
closing because of #459 459

PR #463, opened 2024-07-25
Allowing Project Paths to be passed to scrunch.dataset.fork() method

shaikh-ma:
The `owner` parameter of the `fork()` method allows creating a dataset fork in a defined project location.
But to make it work the `owner` parameter _always_ needs to be the Project URL e.g., `https://<company>.crunch.io/api/projects/<project_id>/`.

This often requires the user, an additional step, to manually find out the project URL to pass as the `owner` parameter.

It is comparatively easier to work with the project folder paths e.g., `"Client datasets | Projects A | Project A - 1 "` than the project URL, which also is more human friendly.
Hence, this Pull Request adds a new feature to `fork()` method to allow the user to pass the project path to the `owner` parameter as well, as requested in the [issue](https://github.com/Crunch-io/scrunch/issues/464).

The previous functionality of the `fork()` method is still the same, only the behavior of the `owner` parameter has been updated.

On passing project path (instead of a project URL), the `fork()` method now automatically maps the folder path to its respective Project URL. Thus, preventing the user an additional step for finding out the Project URL manually.

Examples:
```
import scrunch

ds = scrunch.get_dataset("Dataset X", project="Project A")

# Passing project path to `owner`
ds_fork = ds.fork(preserve_owner=False, owner="Project B | Project 1")


# Passing project URL to `owner`
ds_fork = ds.fork(preserve_owner=False, owner="https://<company>.crunch.io/api/projects/<project_id>/")

```

Additionally, this PR handles the timeout error (`TaskProgressTimeoutError`) which often occurs when forking large datasets.

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2419,7 +2419,11 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
 
         # not returning a dataset
         payload = shoji_entity_wrapper(body)
-        _fork = self.resource.forks.create(payload).refresh()
+        try:
+            _fork = dataset.resource.forks.create(payload).refresh()
+        except TaskProgressTimeoutError as exc:
+            _fork = exc.entity.wait_progress(exc.response).refresh()
+            
         # return a MutableDataset always
         fork_ds = MutableDataset(_fork)  # Fork has same editor as current user
         return fork_ds
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2392,7 +2392,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         :returns _fork: scrunch.datasets.BaseDataset
         """
         from scrunch.mutable_dataset import MutableDataset
-        
+
         description = description or self.resource.body.description
 
         if name is None:
@@ -2408,7 +2408,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
             is_published=is_published,
             **kwargs
         )
-        
+       
         # Handling project vs owner conflict
         owner = kwargs.get("owner")
 
@@ -2432,15 +2432,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
 
         payload = shoji_entity_wrapper(body)
 
-        try:
-            _fork = self.resource.forks.create(payload).refresh()
-        except TaskProgressTimeoutError as exc:
-            # To avoid getting another timeout,
-            # setting timeout=None & increasing the interval.
-            _fork = exc.entity.wait_progress(
-                exc.response,
-                progress_tracker=DefaultProgressTracking(timeout=None, interval=10)
-            ).refresh()
+        _fork = self.resource.forks.create(payload).refresh()
         return MutableDataset(_fork)
 
 
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2367,7 +2367,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         return self.decks[new_deck.self.split('/')[-2]]
 
     def fork(self, description=None, name=None, is_published=False,
-        preserve_owner=True, **kwargs):
+        preserve_owner=True, project=None, **kwargs):
         """
         Create a fork of ds and add virgin savepoint.
 
@@ -2385,13 +2385,15 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
             If True, the owner of the fork will be the same as the parent
             dataset otherwise the owner will be the current user in the
             session and the Dataset will be set under `Personal Project`.
-        :kwargs owner: str (optional)
-            When preserve_owner=False, the owner of fork will be the set
-            according to the passed Project URL or Folder path.
+        :param project: str, default=None
+           The project ID or URL for the project in which the fork 
+           dataset should be created.
 
         :returns _fork: scrunch.datasets.BaseDataset
         """
         from scrunch.mutable_dataset import MutableDataset
+        
+        description = description or self.resource.body.description
 
         if name is None:
             nforks = len(self.resource.forks.index)
@@ -2400,7 +2402,6 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                 dsname = dsname.encode("ascii", "ignore")
             name = "FORK #{} of {}".format(nforks + 1, dsname)
 
-        description = description or self.resource.body.description
 
         body = dict(
             name=name,
@@ -2408,23 +2409,36 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
             is_published=is_published,
             **kwargs
         )
-
+        
+        # Handling project vs owner conflict
         owner = kwargs.get("owner")
-        if preserve_owner:
-            body['owner'] = self.resource.body.owner
+
+        if project and owner:
+            raise ValueError(
+                "Cannot pass both 'project' & 'owner' parameters together. "
+                "Please try again by passing only 'project' parameter."
+                )
         elif owner:
+            project = owner
+        
+        # Setting project value based on presere_owner
+        if preserve_owner and project:
+            raise ValueError("Cannot pass 'project' or 'owner' when preserve_owner=True")
+        elif preserve_owner:
+            body["owner"] = self.resource.body.owner
+        elif project:
             body["owner"] = (
-                owner if owner.startswith("http") else get_project(owner).url
+                project if project.startswith("http") else get_project(project).url
             )
 
-        # not returning a dataset
         payload = shoji_entity_wrapper(body)
+
         try:
             _fork = self.resource.forks.create(payload).refresh()
         except TaskProgressTimeoutError as exc:
             _fork = exc.entity.wait_progress(exc.response).refresh()
-
         return MutableDataset(_fork)
+
 
     def replace_values(self, variables, filter=None, literal_subvar=False, timeout=60):
         """
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2384,7 +2384,10 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         :param preserve_owner: bool, default=True
             If True, the owner of the fork will be the same as the parent
             dataset otherwise the owner will be the current user in the
-            session and the Dataset will be set under `Persona Project`
+            session and the Dataset will be set under `Personal Project`.
+        :kwargs owner: str (optional)
+            When preserve_owner=False, the owner of fork will be the set
+            according to the passed Project URL or Folder path.
 
         :returns _fork: scrunch.datasets.BaseDataset
         """
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2420,13 +2420,11 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
         # not returning a dataset
         payload = shoji_entity_wrapper(body)
         try:
-            _fork = dataset.resource.forks.create(payload).refresh()
+            _fork = self.resource.forks.create(payload).refresh()
         except TaskProgressTimeoutError as exc:
             _fork = exc.entity.wait_progress(exc.response).refresh()
-            
-        # return a MutableDataset always
-        fork_ds = MutableDataset(_fork)  # Fork has same editor as current user
-        return fork_ds
+
+        return MutableDataset(_fork)
 
     def replace_values(self, variables, filter=None, literal_subvar=False, timeout=60):
         """
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -15,11 +15,6 @@ except ImportError:
     pd = None
 
 
-if sys.version_info.major == 3:
-    from collections.abc import Mapping
-else:
-    from collections import Mapping
-
 import six
 
 import pycrunch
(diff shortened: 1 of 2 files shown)


shaikh-ma:
Hello @jjdelc, @aless10, 👋

Please could you take a look at this PR, when you get a chance.


Thanks

jjdelc (changes requested):
The intention of this ticket is correct, it is a good direction to move away from dataset "owners" and think of them as living inside a project.

But I don't think this implementation reflects that.

What I suggest is using a new argument called `project` and have it receive a Project instance (or maybe a project ID).

That way you can call `dataset.fork(project=destination_project)`

And I think that the default behavior should be:
```
def fork(self, ..., project=None, ...):
  project = project or self.project
```

This way, new forks without any location information would be created on the same project as the current dataset (the user being an editor on the dataset to issue a fork, should also have edit permissions on the project to place the new fork)

Bonus points: remove the owner stuff completely, it is not an official use case anymore, is anybody using it?

shaikh-ma:
This sounds a great idea @jjdelc! 🙂

Initially @alextanski & I, had a similar thought to add a new parameter "project".

But we had few concerns about it, then:
1. It might become confusing both the developer & the users when/how to use the `owner` & `project` parameter. 
2. Possibility of undesired scenarios e.g., user might be passing both these parameters or none. Et cetra.
3. The users might need to update their scripts for the new behaviour.

Hence, the current changes were aimed to not change the current behaviour of the `fork()` method while adding the needed features additionally. So as to maintain simplicity & backward compatibility.

As you've now suggested to remove the "owner" completely, just a few questions:
1. How should we handle the scenario when the user needs the fork dataset to be created in their personal project?
2. Should we rename or remove the `preserve_owner` parameter?

Thanks

jjdelc:
You make a good point about backwards compatibility, and I think it should be easy to handle in inside the method.

For now, you may need to keep the `owner` behavior as it is. And add the new `project` field. Handle inside the call if both are sent raise an error and suggest to only send project, if none are sent, follow the current behavior.

My other point is not to require the user to send URLs, but allow for a Project() instance to be received, I would like to support arbitrary paths too, but you'll risk it that the path may not exist and have to then take care of creating and such, best to have that responsibility outside of this `.fork()` call and once they solved that they would have a Project() instance.

Also, add a test checking your payload.

scrunch/datasets.py
```diff
@@ -2412,14 +2409,22 @@ def fork(self, description=None, name=None, is_published=False,
             **kwargs
         )
 
+        owner = kwargs.get("owner")
         if preserve_owner:
             body['owner'] = self.resource.body.owner
+        elif owner:
+            body["owner"] = (
+                owner if owner.startswith("http") else get_project(owner).url
```
jjdelc, line 2417: These should be encouraged to be Crunch objects rather than strings. For projects don't allow strings, urls or paths. Because you'll have problems validating here.

Have this be a Project instance.

scrunch/datasets.py
```diff
@@ -2412,14 +2409,22 @@ def fork(self, description=None, name=None, is_published=False,
             **kwargs
         )
 
+        owner = kwargs.get("owner")
         if preserve_owner:
             body['owner'] = self.resource.body.owner
```
jjdelc, line 2414: If they choose to preserve owner, what they mean is that it should be in the same place as the current project. The API already returns the project URL for the `owner` attribute. So this may also be `self.resource.project`.

And we can slowly stop the conversaion about owners, because that concept does not exist in the API anymore and will bring confusion to users that individuals have ownership of datasets.

scrunch/datasets.py
```diff
@@ -2412,14 +2409,22 @@ def fork(self, description=None, name=None, is_published=False,
...
         # not returning a dataset
         payload = shoji_entity_wrapper(body)
-        _fork = self.resource.forks.create(payload).refresh()
-        # return a MutableDataset always
-        fork_ds = MutableDataset(_fork)  # Fork has same editor as current user
-        return fork_ds
+        try:
+            _fork = self.resource.forks.create(payload).refresh()
+        except TaskProgressTimeoutError as exc:
+            _fork = exc.entity.wait_progress(exc.response).refresh()
```
jjdelc, line 2425: If you get a timeout,  you should continue to raise, this is an invalid call with the exception. If it timed out it quite likely failed, and the code flow shouldn't continue bc you won't have a fork to carry on with.

jjdelc (changes requested):

scrunch/datasets.py
alexbuchhammer: @jjdelc **question:** Do you refer here to catching the timeout inside the fork() method directly? Because it is common practice for users to catch the timeouts on forks with a [progress tracking](https://github.com/Crunch-io/pycrunch/blob/master/src/pycrunch/progress.py#L21) against a larger timeout. We simply do not want to have this inside the method here?

scrunch/datasets.py
alexbuchhammer: **question:** @jjdelc I am a bit confused here - We want to keep the `preserve_owner` kwarg and add a new one, `project`, explicitly to the signature by your suggestion; then basically fall back to `self.resource.project` (simple code change to abandon the old  `owner` property) if `project` is not given?

scrunch/datasets.py
alexbuchhammer: **question:** @jjdelc Double-checking on this - Isn't [get_project()](https://github.com/Crunch-io/scrunch/blob/master/scrunch/datasets.py#L180) validating the string for us? It will return a project instance if the path can be resolved, and raise an `InvalidPathError` it cannot:

```
InvalidPathError                          Traceback (most recent call last)
Cell In[8], line 7
      4 ygdc_crunch.Crunch()
      6 name_pr = "TechOps|Alex 3"
----> 7 pr = scrunch.get_project(name_pr)

File ~\AppData\Local\Continuum\anaconda2\envs\py311test\Lib\site-packages\scrunch\datasets.py:206, in get_project(project, connection)
    203 _project = Project(ret)
    205 if sub_project:
--> 206     _project = _project.get(sub_project)
    208 return _project

File ~\AppData\Local\Continuum\anaconda2\envs\py311test\Lib\site-packages\scrunch\datasets.py:552, in Project.get(self, path)
    550 for p_name in Path(path).get_parts():
    551     try:
--> 552         node = node.get_child(p_name)
    553     except KeyError:
    554         raise InvalidPathError('Project not found %s' % p_name)

File ~\AppData\Local\Continuum\anaconda2\envs\py311test\Lib\site-packages\scrunch\datasets.py:571, in Project.get_child(self, name)
    568         return Project(tup.entity)
    569     return self.root.dataset[name]
--> 571 raise InvalidPathError('Invalid path: %s' % name)

InvalidPathError: Invalid path: Alex 3
```

scrunch/datasets.py
jjdelc: Yes, really at this point owner and project are equivalent. There is no concept of an individual owner in Crunch.

It looks like `preserve_owner` is the default behavior when no owner is passed in, if preserve_owner is False, and no owner is passed, then there's no explicit code path for this.

scrunch/datasets.py
jjdelc: What I mean is that if pycrunch raised a timeout, there's no recovery. Calling `.wait_progress` with the response of the timeout is not going to work, and it won't achieve anything (maybe some python error) but this line looks like it is attempting some fallback way to recover from the timeout and still obtain a `_fork` instance from it.

alexbuchhammer: Yep - got it now, @jjdelc! Make sense. @shaikh-ma I think if we would want control the timeout AND also return the fork when it completed, we would need to do this:

```
... < catch & wait >
entity = exc.entity.refresh()
fork_ds = self.get_dataset(entity.body.id)
```

I think this _should_ work?

scrunch/datasets.py
alexbuchhammer: Thanks @jjdelc - so a green light on allowing string paths?

scrunch/datasets.py
alexbuchhammer: Thanks for confirming.

scrunch/datasets.py
shaikh-ma: The line `_fork = exc.entity.wait_progress(exc.response).refresh()` worked as expected (without giving any error) when I tried out with a dataset. 

Handling TimeOut error

scrunch/tests/test_datasets.py
```diff
@@ -1777,15 +1777,64 @@ def _create(*args):
...
+            "Cannot pass both 'project' & 'owner' parameters together. "
+            "Please try again by passing only 'project' parameter."
+        )
+        with pytest.raises(ValueError, match=err_msg1):
+            ds.fork(owner="ABCD", project="1234")
+        
+        err_msg2 = ("Cannot pass 'project' or 'owner' when preserve_owner=True")
+        with pytest.raises(ValueError, match=err_msg2):
+            ds.fork(preserve_owner=True, project="1234")
+            ds.fork(preserve_owner=True, owner="1234")
```
aless10, line 1822: @shaikh-ma I am not sure about this test. Could you try to rewrite it like this?

```suggestion
        err_msg2 = ("Cannot pass 'project' or 'owner' when preserve_owner=True")
        with pytest.raises(ValueError, match=err_msg2):
            ds.fork(preserve_owner=True, project="1234")
            
         with pytest.raises(ValueError, match=err_msg2):
            ds.fork(preserve_owner=True, owner="1234")
```

so that you can test the different scenatios? I am just not sure that the check you wrote 
```
# Setting project value based on presere_owner
        if preserve_owner and project:
            raise ValueError("Cannot pass 'project' or 'owner' when preserve_owner=True")
```
does what expected. But if the test passes, then it's good. Thanks

shaikh-ma: Done

scrunch/datasets.py
jjdelc: So this looks like it is continuing to wait after the client has timed out, say pycrunch's timeout is 2m and the operation is taking longer than that.

On the first exception the client will raise TimeOut, but the server may still carry on updating the progress. So that's what that 2nd attempt is doing, it's giving it another chance to continue. That 2nd one on its own may also raise Timeout in case it takes longer than, now, 4 minutes. And you'll be back at the same situation.

This is equivalent of saying `timeout = 2*default_timeout`. You've effectively waited 2 times. 

It's a way to get around longer tasks, but code wise it reads very strange as it is not clear what the intention of this 2nd attempt is. Maybe add some comments, or even better if you want to increase the timeout then best to configure a longer timeout all together

scrunch/tests/test_datasets.py
```diff
@@ -1777,15 +1777,66 @@ def _create(*args):
...
     def test_fork_preserve_owner(self):
         user_id = 'http://test.crunch.io/api/users/123/'
+        project_id = 'http://test.crunch.io/api/projects/123/'
         sess = MagicMock()
         body = JSONObject({
             'name': 'ds name',
             'description': 'ds description',
-            'owner': user_id,
+            'owner': project_id,
+            # 'owner': user_id,
```
jjdelc, line 1836: This comment needs to go?

scrunch/tests/test_datasets.py
```diff
@@ -1795,7 +1846,8 @@ def test_fork_preserve_owner(self):
             'body': {
                 'name': 'FORK #1 of ds name',
                 'description': 'ds description',
-                'owner': user_id,  # Owner preserved
+                # 'owner': user_id,  # Owner preserved
+                'owner': project_id,
```
jjdelc, line 1850: I think it's best to start updating the tests to use `project` to make sure anybody else following the implementation starts learning this is the preferred way

scrunch/tests/test_datasets.py
shaikh-ma: I've removed the comment

scrunch/tests/test_datasets.py
shaikh-ma: Updated

shaikh-ma: Hi @jjdelc ,

I've updated the `try-catch` block to set timeout=None to avoid getting timed out again & have also added a comment stating the purpose.

jjdelc: I continue to worry that this re-attempt strategy is a bad approach. What this is doing is capturing the 1st timeout and then trying again. 

Recall that this is a client-side timeout, say 60s so in case the opreation in the backend ends after 61s, the client will time out, and raise an error, then the client will try again a 2nd time and proceed for a next creation.

I think that if the issue is to reduce timeout changes, this should simply set a long timeout on the very fist request. Instead of making 2 independent creation requests to the server.

shaikh-ma: Hi @jjdelc,

As we've already completed the main goal of this PR, which involves allowing Project paths & URL, 
I think it would be a good idea to explore timeout handling separately. 

Therefore, I've removed the timeout handling part from the `fork()` method as this step is just an extra precaution to handle timeouts, but it's not the primary focus of this PR.

jjdelc: Very good call @shaikh-ma , I think this is best, if the fork starts to timeout, we can address that later.

jjdelc (approved):

aless10 (approved):

shaikh-ma:
Hi @jjdelc 👋 ,

Please let me know if anything else is needed from our end or it's good to be merged. 😄

jjdelc:
Merged on master, I'll make another minor release to include this

PR #465, opened 2024-07-31
Add strict subvariable syntax handler

aless10:
Add `strict_subvariable_syntax` flag handler for system script + tests

### DIFF
--- a/scrunch/accounts.py
+++ b/scrunch/accounts.py
@@ -2,7 +2,7 @@
 
 import pycrunch
 from scrunch.helpers import shoji_view_wrapper
-from scrunch.scripts import ScriptExecutionError
+from scrunch.scripts import ScriptExecutionError, SystemScript
 from scrunch.connections import _default_connection
 from scrunch.datasets import Project
 
@@ -50,20 +50,11 @@ class Account:
         act_res = site_root.account
         return cls(act_res)
 
-    def execute(self, script_body):
-        """
-        Will run a system script on this account.
-
-        System scripts do not have a return value. If they execute correctly
-        they'll finish silently. Otherwise an error will raise.
-        """
+    def execute(self, script_body, strict_subvariable_syntax=None):
+        """Will run a system script on this account."""
         # The account execution endpoint is a shoji:view
-        payload = shoji_view_wrapper(script_body)
-        try:
-            self.resource.execute.post(payload)
-        except pycrunch.ClientError as err:
-            resolutions = err.args[2]["resolutions"]
-            raise ScriptExecutionError(err, resolutions)
+        system_script = SystemScript(self.resource)
+        system_script.execute(script_body, strict_subvariable_syntax)
 
     @property
     def projects(self):
--- a/scrunch/accounts.py
+++ b/scrunch/accounts.py
@@ -54,7 +54,7 @@ class Account:
         """Will run a system script on this account."""
         # The account execution endpoint is a shoji:view
         system_script = SystemScript(self.resource)
-        system_script.execute(script_body, strict_subvariable_syntax)
+        return system_script.execute(script_body, strict_subvariable_syntax)
 
     @property
     def projects(self):
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -33,14 +33,13 @@ from scrunch.exceptions import InvalidParamError, InvalidVariableTypeError
 from scrunch.expressions import parse_expr, prettify, process_expr
 from scrunch.folders import DatasetFolders
 from scrunch.views import DatasetViews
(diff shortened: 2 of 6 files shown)


scrunch/accounts.py
```diff
@@ -50,20 +50,11 @@ def current_account(cls, connection=None):
...
+    def execute(self, script_body, strict_subvariable_syntax=None):
+        """Will run a system script on this account."""
         # The account execution endpoint is a shoji:view
-        payload = shoji_view_wrapper(script_body)
-        try:
-            self.resource.execute.post(payload)
-        except pycrunch.ClientError as err:
-            resolutions = err.args[2]["resolutions"]
-            raise ScriptExecutionError(err, resolutions)
+        system_script = SystemScript(self.resource)
```
jjdelc, line 56: 👍🏽  Good abstraction

scrunch/datasets.py
```diff
@@ -2224,7 +2209,7 @@ def export(self, path, format='csv', filter=None, variables=None,
                     % (format, ','.join(k))
                 )
         if 'var_label_field' in options \
-                and not options['var_label_field'] in ('name', 'description'):
+                and options['var_label_field'] not in ('name', 'description'):
```
jjdelc, line 2212: 👍🏽

scrunch/scripts.py
```diff
@@ -44,19 +46,50 @@ def get_default_syntax_flag(self, strict_subvariable_syntax):
...
+    def execute(self, script_body, strict_subvariable_syntax=None):
+        pass
+
+
+class SystemScript(BaseScript):
+
+    def format_request_url(self, execute, strict_subvariable_syntax=None):
+        strict_subvariable_syntax_flag = self.get_default_syntax_flag(strict_subvariable_syntax)
+        if strict_subvariable_syntax_flag:
+            execute.self += "?strict_subvariable_syntax=true"
```
jjdelc, line 61: Here you will be mutating the `self` attribute of the `execute` resource. I think that this better returns the desired URL instead of altering the instance.

Because when you execute a script for the 2nd time this will add again this string incorrectly

scrunch/accounts.py
```diff
@@ -50,20 +50,11 @@ def current_account(cls, connection=None):
...
+        system_script.execute(script_body, strict_subvariable_syntax)
```
jjdelc, line 57: I see that the original wasn't returning anything, should we return the Request obejct? This would allow the caller to get a hold of the progress URL and track it, otherwise the return is None and it's impossible for the caller to know where to track this progress

scrunch/scripts.py
```diff
@@ -44,19 +46,50 @@ def get_default_syntax_flag(self, strict_subvariable_syntax):
...
+
+        System scripts do not have a return value. If they execute correctly
+        they'll finish silently. Otherwise, an error will raise.
+        """
+        # The project execution endpoint is a shoji:view
+        payload = shoji_view_wrapper(script_body)
+        try:
+            execute = self.resource.execute
+            self.format_request_url(execute, strict_subvariable_syntax)
+            return execute.post(payload)
```
jjdelc, line 75: I see why you had to mutate the instance, I think that here it's best to use a raw request, see how this pycrunch `.post()` is implemented and you'll do a `self.resource.session.post(execute_url. ....)`.

jjdelc (changes requested):

scrunch/accounts.py
aless10: I did that for the projects because I added an integration test, not for this one. Thanks

scrunch/scripts.py
aless10: That is not completely true. I am not mutating the self.resource instance, but just the local `execute`.
Every time you do `self.resource.something` there is a GET call to the `resource` o get the attribute. 
So the self.resource instance is not modified by this, I am just updating the local object that is the result of the GET
So here https://github.com/Crunch-io/scrunch/blob/86b3b7337d498c057c8f1af443d5aa056cc1073e/scrunch/scripts.py#L73-L73 I am fetching the `execute` object, and I am modifing this one that is the result of this call https://github.com/Crunch-io/pycrunch/blob/9bb073368293f1e84843a53bdc0cfccd1c90787e/src/pycrunch/elements.py#L154
Next time, next script, I am going to redo the request to the same self.resource object, that has not being modified.
Let me know if that works for you. Otherwise I'll think of another way of doing this. But I just wanted to use the pycrunch interface as per the DatasetsScripts

scrunch/scripts.py
aless10: As explained above, I did not mutate the instance, so we should be safe

PR #470, opened 2024-10-16
Removing version pinning for requests library

shaikh-ma:
In this PR, we've removed the version pinning for `request` library in order to be able to use it's latest version.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ params = dict(
     python_requires='>=2.7',
     install_requires=[
         "pycrunch==0.5.5",
-        "requests==2.27.0",
+        "requests",
         crunch_cube,
         'six',
     ],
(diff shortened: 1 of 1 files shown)


alexbuchhammer:
@jjdelc Hey - can you please take a look and let us know if this could be merged? We are currently working hard to roll-out a full python 3.11 Data Processing infrastructure for the YG CENX and the pinned `requests` lib version is causing some compatibility issues.

setup.py
```diff
@@ -47,7 +47,7 @@
     python_requires='>=2.7',
     install_requires=[
         "pycrunch==0.5.5",
-        "requests==2.27.0",
```
jjdelc, line 50: Can you use the `if PY27` to use the pinned version and use unpinned otherwise? I am a tad concerned that this may make python2.7 installations that YG need problematic.

jjdelc (changes requested):

setup.py
shaikh-ma: Great Idea @jjdelc! 😃

I've updated as per your feedback.

jjdelc (approved):

PR #473, opened 2024-11-26
Task/update fork method preserve owner parameter

shaikh-ma:
As informed by Crunch team, the special properties of the "Personal" folder are about to be replaced with a regular project folder that behaves like any other Crunch folder, at some point approaching the end of CY 2024. 

Recently added functionality to create a fork in a target location will still allow to use `preserve_owner` and hence might lead to creating the fork in either the "Personal" folder or keeping the source dataset's location. 

This MR proactively attempts to push towards the best-practice approach here by updating the fork method.
The behaviour of the preserve_owner parameter has been updated as described below:

- If `preserve_owner=False` and a project is not provided, we want to raise and indicate that an explicit project location is expected.
- If `preserve_owner=False` and a project is provided, this is consistent with what we want to acheive and the code can be executed as is.
- If `preserve_owner=True` and a project is provided, raise and indicate that this combination is ambiguous, only one of the two can be used AND that preserve_owner will be deprecated soon.
- Regardless of the value given to `preserve_owner`, log a `DeprecationWarning` that the `preserve_owner` parameter will be removed soon in favour of providing a project or not (in which case the behavior will be as it is currently if `preserve_owner=True and project=None`).

|Project passed|Preserve_owner = True|
|-|-|
|Yes| image|
|No| image|

|Project passed| Preserve_owner = False|
|-|-|
|Yes| image|
|No| image|

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -2379,10 +2379,19 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
 
         :returns _fork: scrunch.datasets.BaseDataset
         """
-
         from scrunch.mutable_dataset import MutableDataset
+        
+        # Handling project vs owner conflict
+        owner = kwargs.get("owner")
 
-        description = description or self.resource.body.description
+        if project and owner:
+            raise ValueError(
+                "Cannot pass both 'project' & 'owner' parameters together. "
+                "Please try again by passing only 'project' parameter."
+        )
+        elif owner:
+            project = owner
+            del kwargs["owner"]
 
         LOG.warning(
             "The preserve_owner parameter will be removed soon"
@@ -2400,22 +2409,10 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
 
         body = dict(
             name=name,
-            description=description,
+            description=description or self.resource.body.description,
             is_published=is_published,
             **kwargs
         )
-
-        # Handling project vs owner conflict
-        owner = kwargs.get("owner")
-
-        if project and owner:
-            raise ValueError(
-                "Cannot pass both 'project' & 'owner' parameters together. "
-                "Please try again by passing only 'project' parameter."
-        )
-        elif owner:
-            project = owner
-
         # Setting project value based on preserve_owner.
         if preserve_owner:
             if project:
@@ -2424,7 +2421,7 @@ class BaseDataset(ReadOnly, DatasetVariablesMixin):
                 )
             else:
                 # Create fork in source dataset path.
-                body["project"] = self.resource.body.owner
+                body["project"] = self.resource.body.owner 
         else:
             if project:
                 # Create fork in given Project path.
@@ -3510,5 +3507,5 @@ class BackfillFromCSV:
                 if folder_name in folders_by_name:
                     folders_by_name[folder_name].entity.delete()
                 # Always delete the tmp dataset no matter what
-                tmp_ds.delete()
+                tmp_ds.delete)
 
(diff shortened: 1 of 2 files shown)


alexbuchhammer:
Hey @jjdelc - what's your opinion on the suggested changes? We do not yet want to get rid of the preserve_owner parameter to protect users from breaking scripts. Let us know!

Cheers
Alex

jjdelc:
This looks in the right direction, but I worry that the change may come sooner than expected. I will check with @rchacon1  on what the status of this is in the backend.

scrunch/datasets.py
```diff
@@ -2371,15 +2371,23 @@ def fork(self, description=None, name=None, is_published=False,
...
         :returns _fork: scrunch.datasets.BaseDataset
         """
+
         from scrunch.mutable_dataset import MutableDataset
 
         description = description or self.resource.body.description
 
+        LOG.warning(
+            "The preserve_owner parameter will be removed soon"
+            " in favour of providing a project or not "
```
rchacon1, line 2386: A `project` will always be required as far the API is concerned.

jamesrkg: The idea he is, as an abstraction over the API, the user can either give a project or not, and if they don't, then the fork will be created in the same location as the dataset being forked. What happens below the signature of the method isn't something we want users to think about. In this way, it will be possible for a user to say `ds.fork(..., project=None)`, even though that's not what is implemented internally.

scrunch/datasets.py
```diff
@@ -2371,15 +2371,23 @@ def fork(self, description=None, name=None, is_published=False,
             dataset otherwise the owner will be the current user in the
             session and the Dataset will be set under `Personal Project`.
         :param project: str, default=None
-           The project ID or URL for the project in which the fork 
-           dataset should be created.
+            The project ID or URL for the project in which the fork
```
jamesrkg, line 2374: To make the behavior of project clearer, we should add to the docstring an explanation of what happens when project is None.

shaikh-ma: I've updated the docstring for the `project` & `preserve_owner` parameters.

scrunch/datasets.py
rchacon1: if you want to keep the `preserve_owner` abstraction you can future proof it by sending `resource.body.owner` as the `project` in the payload.

scrunch/datasets.py
```diff
@@ -2393,27 +2404,37 @@ def fork(self, description=None, name=None, is_published=False,
...
-                project if project.startswith("http") else get_project(project).url
-            )
+        if preserve_owner:
+            if project:
+                raise ValueError(
+                    "Cannot pass 'project' or 'owner' when preserve_owner=True."
+                )
+            else:
+                # Create fork in source dataset path.
+                body["owner"] = self.resource.body.owner
```
rchacon1, line 2427: you can replace `body["owner"]` with `body["project"]` to future proof this

scrunch/datasets.py
jamesrkg: Thanks, yep that's what happening here:

https://github.com/Crunch-io/scrunch/blob/930905a298dc92b9f02497b59f1da1de508dd934/scrunch/datasets.py#L2420-L2427

scrunch/datasets.py
jamesrkg: Thanks @shaikh-ma 🙇

scrunch/datasets.py
shaikh-ma: Done

rchacon1 (approved):
:shipit:

shaikh-ma:
Hi @rchacon1 @jjdelc,

Please could you push these changes in a tag release. 🙇

rchacon1:
version 0.18.6 published to pypi

PR #476, opened 2024-12-19
Update integration tests to include project in payload

rchacon1:
a temp folder for temp ds creation. Also, update integration tests

https://www.pivotaltracker.com/n/projects/2172644/stories/188703975

### DIFF
--- a/scrunch/datasets.py
+++ b/scrunch/datasets.py
@@ -3391,14 +3391,7 @@ class BackfillFromCSV:
                 schema[a] = svdef
         return schema
 
-    def make_tmp_name(self):
-        return "Scrunch-backfill-{}-{}-{}".format(
-            self.dataset.name, self.dataset.id, self.timestamp)
-
-    def create_tmp_project(self, name):
-        return self.root.projects.create(shoji_entity_wrapper({"name": name}))
-
-    def create_tmp_ds(self, name, project, csv_file):
+    def create_tmp_ds(self, csv_file):
         """
         Creates a new pycrunch dataset:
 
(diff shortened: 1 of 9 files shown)


scrunch/datasets.py
```diff
@@ -3391,26 +3391,31 @@ def obtain_schema(self):
...
 
-    def create_tmp_ds(self, csv_file):
+    def make_tmp_name(self):
+        return "Scrunch-backfill-{}-{}-{}".format(
+            self.dataset.name, self.dataset.id, self.timestamp)
+
+    def create_tmp_project(self, name):
+        return self.root.projects.create(shoji_entity_wrapper({"name": name}))
+
+    def create_tmp_ds(self, name, project, csv_file):
```
rchacon1, line 3401: @jjdelc I think this change might require us to bump this version to 0.19.0. What do you think?

scrunch/datasets.py
```diff
@@ -3508,4 +3515,3 @@ def execute(self, csv_file):
                     folders_by_name[folder_name].entity.delete()
                 # Always delete the tmp dataset no matter what
                 tmp_ds.delete()
```
rchacon1, line 3517: If we delete `tmp_project` here we run into the issue where `DELETE /projecsts/<project_id>` moves expired datasets to the personal folder. @jjdelc what do you think we should do here?

scrunch/datasets.py
```diff
@@ -3410,7 +3410,8 @@ def create_tmp_ds(self, csv_file):
             "table": {
                 "element": "crunch:table",
                 "metadata": metadata
-            }
+            },
+            "project": self.dataset.project.url,
```
rchacon1, line 3414: @jjdelc you think it's okay to use the same project as the source dataset?

jjdelc: Yes absolutely, because it ensures you have write permissions.

jjdelc (approved):

PR #478, opened 2025-01-17
Updates to fork function for allowing Projects ID instead of URLs

shaikh-ma:


alexbuchhammer:
Thanks @shaikh-ma - this looks like a great quality of life change.

@jjdelc @rchacon1 - what do you guys think? Do you see any issues with this change?

Cheers
Alex

scrunch/datasets.py
```diff
@@ -2425,9 +2425,15 @@ def fork(self, description=None, name=None, is_published=False,
...
                 # Create fork in given Project path.
-                body["project"] = (
-                    project if project.startswith("http") else get_project(project).url
-                )
+                try:
+                    project = get_project(project).url
+                except KeyError:
+                    # Creating full project URL for sub-folders
+                    connection = _default_connection(connection=None)
+                    site_url = connection.session.site_url
```
jjdelc, line 2433: You can obtain this from `self.resource.session.site_url` without having to make a new connection.

scrunch/datasets.py
```diff
@@ -2425,9 +2425,15 @@ def fork(self, description=None, name=None, is_published=False,
         else:
             if project:
                 # Create fork in given Project path.
-                body["project"] = (
-                    project if project.startswith("http") else get_project(project).url
-                )
+                try:
+                    project = get_project(project).url
```
jjdelc, line 2429:
```suggestion
                    project_url = get_project(project).url
```
It will help to not overwrite the variable and also naming it `project_url` is more clear that this is storing a URL (string) and not a `Project()` instance. Ideally we'd do `project_url:str = ....` but python 2 😒

jjdelc (commented):
Looks good! I just made a couple of minor suggestions

scrunch/datasets.py
shaikh-ma: Thanks for the suggestion, I've updated this. 😃

scrunch/datasets.py
shaikh-ma: Agreed! 😃

alexbuchhammer:
@jjdelc @shaikh-ma - hey guys, seems like this is then good to go? Can we add it to a next possible release, then we can move the users over to that change (which looks much cleaner!). Thank you very much!

Alex

jjdelc (approved):

shaikh-ma:
Hi @jjdelc , 

Please could you add these updates in a new tag release. 🙇

alexbuchhammer:
Ah - I think - if I am recalling some previous discussions in Slack correctly - at least the the platform team is using scrunch as well for streaming/audits/dataset creation, so we should maybe give them a heads-up before that change would be pulled in via a new version release. At least worth checking to avoid prod issues in YG's core data operations. 

@jamesrkg ^^

PR #483, opened 2025-02-07
Cadl 41 py311 integration 2

angelo-romano:


dhandy2013 (approved):
Angelo assured me that he tested this branch locally using the same commands that Circle CI would run, and it passes locally. So we are hopeful that we have found and fixed the cause of the problem that broke branch builds earlier.

PR #492, opened 2025-08-20
Use weight_alias on dataset settings and preferences instead of weight

tiagocoutinho:


scrunch/datasets.py
```diff
@@ -2087,7 +2087,8 @@ def create_crunchbox(self, title='', header='', footer='', notes='',
         if weight is DefaultWeight:
             preferences = self.resource.session.get(
                 self.resource.fragments.preferences)
-            weight = preferences.payload.body.weight or None
+            weight_alias = getattr(preferences.payload.body, "weight_alias", None)
+            weight = None if weight_alias is None else self[weight_alias].url
```
jjdelc, line 2091: If you only need the URL, then doing the `__getitem__` will make an HTTP GET request to the variable entity fetching the full shoji entity only to fetch the URL that we had.

This could be done this way:
```
weight_url = self._catalog.by('alias')[weight_alias].entity_url
```

jjdelc: But my question then would be, why does Scrunch needs this URL, why can it not continue with the alias. How is this being used, could we send the alias in that next request?

scrunch/datasets.py
```diff
@@ -2087,7 +2087,8 @@ def create_crunchbox(self, title='', header='', footer='', notes='',
         if weight is DefaultWeight:
             preferences = self.resource.session.get(
                 self.resource.fragments.preferences)
-            weight = preferences.payload.body.weight or None
+            weight_alias = getattr(preferences.payload.body, "weight_alias", None)
```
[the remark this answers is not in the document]
tiagocoutinho: I tried:
```python
weight_alias = preferences.payload.body.get("weight_alias")
```
but I get:
```
AttributeError: 'dict_to_obj' object has no attribute 'get'
```
that is why I used `getattr`. I saw other places where the same pattern is used

scrunch/datasets.py
tiagocoutinho: Looks like it should be fine to do `self[weight_alias].url` because `__getitem__` already does what you suggest:
```

    def __getitem__(self, item):
        variable = self._catalog.by('alias').get(item)
        if variable is None:  # Not found by alias
            ...
        return Variable(variable, self)
```

jjdelc (approved):

PR #498, opened 2026-05-08
fixes for deprecated project.datasets endpoint.

percious:


scrunch/tests/test_datasets.py
```diff
@@ -3868,7 +3868,7 @@ def setUp(self):
 
         proj_resource = MagicMock()
         proj_resource.self = 'http://test.crunch.local/api/projects/42/'
-        proj_resource.datasets = datasets
+        proj_resource.index = datasets
```
jjdelc, line 3871: This isn't quite right, because it will mock the wrong pycrunch API. The index should be a dictionary keyed by URL. This `datasets` object is emulating a ShojiCatalog instance and not an Index.

scrunch/datasets.py
```diff
@@ -510,12 +510,12 @@ def edit_user(self, user, edit):
         self.members.edit(user, {'permissions': {'edit': edit}})
 
     def get_dataset(self, dataset):
-        datasets = self.resource.datasets
+        index = self.resource.index
         try:
-            shoji_ds = datasets.by('name')[dataset].entity
+            shoji_ds = index.by('name')[dataset].entity
```
jjdelc, line 515: Yup, this will break. The change here should be 
```suggestion
            shoji_ds = self.resource.by('name')[dataset].entity
```

`index` does not have a `.by` method.

scrunch/datasets.py
```diff
@@ -510,12 +510,12 @@ def edit_user(self, user, edit):
         self.members.edit(user, {'permissions': {'edit': edit}})
 
     def get_dataset(self, dataset):
-        datasets = self.resource.datasets
+        index = self.resource.index
```
jjdelc, line 513:
```suggestion
        resource = self.resource
```
And use `resource` instead of `datasets`

jjdelc (changes requested):

Issue #503, opened 2026-06-22
[BUG] KeyError 'entity' when using add_filter function in scrunch v0.18.23


shaikh-ma:
### Bug Report

**Scrunch Version:** 0.18.23
**Python Version:** 3.11
**OS:** Windows
**Last Working Version:** < 0.18.23

---

### Description

A `KeyError: 'entity'` is raised when attempting to add a filter using the `add_filter` function that involves a multiple response variable. This is a **regression** introduced in version **0.18.23** — the same operation works without issue in prior versions.

The error occurs in `expressions.py` within the `_update_values_for_multiple_response` function, where the code expects `vars_by_alias[var_alias]` to contain an `"entity"` key, but it doesn't.

---

### Traceback

```python
File c:\users\..\scrunch\scrunch\expressions.py:541, in _update_values_for_multiple_response(new_values, values, subitem, vars_by_alias, arrays)
    537     values[0]["column"] = value
    538 values[0].pop("value", None)
    539 subvar_ids_by_aliases = {
    540     v["alias"]: k
--> 541     for k, v in vars_by_alias[var_alias]["entity"]["subvariables"][
    542         "index"
    543     ].items()
    544 }
    545 arrays[0] = [
    546     subvar_ids_by_aliases[new_value["axes"][0]] for new_value in new_values
    547 ]

KeyError: 'entity'
```

---

### Steps to Reproduce

1. Install scrunch v0.18.23:
   ```bash
   pip install scrunch==0.18.23
   ```

2. Connect to a Crunch dataset containing a multiple response variable.

3. Attempt to create a filter referencing that variable:
   ```python
   from scrunch import connect, get_dataset

   site = connect(api_key="YOUR_API_KEY", site_url="https://app.crunch.io/api/")
   ds = get_dataset("Your Dataset Name")
   ds.add_filter(
       name="Test Filter",
       expr="<multiple_response_var> in [1,2]"
   )
   ```

4. Observe the `KeyError: 'entity'` exception.

---

### Expected Behavior

The filter should be created and applied successfully, as it did in all versions prior to 0.18.23.

---

### Actual Behavior

A `KeyError: 'entity'` is raised because `vars_by_alias[var_alias]` no longer contains the `"entity"` key in its dictionary structure.

---

### Workaround

Downgrade to the last known working version:

```bash
pip install scrunch<0.18.23
```

---

### Suggested Fix

[...]

PR #504, opened 2026-07-13
Fix multiple response filters without entity metadata

slobodan-ilic:
## Summary
- support multiple-response variable metadata that is exposed without an entity key
- use table-derived metadata for multiple-response expansion so sparse variables.index metadata still works
- keep scalar subvariable equality untouched while expanding list-style MR filters
- add a regression test for the add_filter-style metadata shape from #503

### DIFF
--- a/scrunch/expressions.py
+++ b/scrunch/expressions.py
@@ -490,19 +490,19 @@ def get_dataset_variables(ds):
 
 
 def get_subvariables_resource(var_url, var_index):
-    variable = var_index[var_url].entity
-    sub_variables = variable.subvariables['index']
-    return {sv['alias'].strip('#'): sv['id'] for sv in sub_variables.values()}
-
-
-def _get_categories_from_var_index(vars_by_alias, var_alias):
-    return vars_by_alias[var_alias].entity.body.categories
+    """Return subvariable alias -> id from table metadata subreferences."""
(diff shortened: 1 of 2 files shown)


scrunch/expressions.py
```diff
@@ -490,19 +490,19 @@ def get_dataset_variables(ds):
...
 def get_subvariables_resource(var_url, var_index):
-    variable = var_index[var_url].entity
-    sub_variables = variable.subvariables['index']
-    return {sv['alias'].strip('#'): sv['id'] for sv in sub_variables.values()}
-
-
-def _get_categories_from_var_index(vars_by_alias, var_alias):
-    return vars_by_alias[var_alias].entity.body.categories
+    """Return subvariable alias -> id from table metadata subreferences."""
+    subreferences = var_index[var_url].get("subreferences", {})
```
jjdelc:
```suggestion
    subreferences = var_index[var_url].get("subreferences", {}) or {}
```

I've found this pattern bits me when the key exists and the value is None. In this case it's going to be super duper weird for this to happen though.

jjdelc (approved):

scrunch/expressions.py
slobodan-ilic: fixed, thanks!

